PR #1, opened 2019-07-22
BUILD: Create a build step for browser extension files

mooeypoo:
The files can now be written independently in `src/` where they
can also be unit tested. A build step `grunt build` now concatenates
the files and inserts them into a browser-extension template for
the content script.
    
Please note:
* All `src/` files should be wrapped with IIFE `(function(){...}())`. This IIFE will be taken out when concatenated.
* If the IIFE needs to be kept through the concatenation process, add a comment to its top (ex `( function () { // don't remove`)
* Files should be added to the Gruntfile.js for the concatenation to continue. All files must be in order.

See https://phabricator.wikimedia.org/T227522


MusikAnimal:
I get `Npm module "grunt-concat-with-template" not found` when running `./node_modules/.bin/grunt build`. Is this expected?

I also see a mix of spacing / tabs, etc. in Gruntfile.js and contentScript.js, though eslint passes for everything in src/, so maybe we don't care :)

Finally, not really an issue with this PR, but we might want to add a .nvmrc. I got a different package-lock.json after running `npm install`.

mooeypoo:
Whoops, I forgot to add `package.json` with the package! fixing that now.

I don't see the mix of spacing/tabs in Gruntfile.js, but I fixed the `contentScript.js` file.

mooeypoo:
And I added a .nvmrc file with node `11.5`.

build/template_browserextension.js
```diff
@@ -0,0 +1,19 @@
+( function ( $, mw ) {
+	function loadWhoWroteThat() {
+		var $button = $( <a> )
```
MusikAnimal: Don't we need quotes around the markup?

extension/manifest.json
```diff
@@ -12,8 +12,10 @@
   "content_scripts": [ {
     "matches": ["*://*.wikipedia.org/*"],
     "all_frames": true,
-    "js": [ "js/lib/jquery.js", "js/content.js" ]
+    "js": [ "js/lib/jquery.js", "js/contentScript.js" ],
+	"run_at": "document_end"
```
MusikAnimal: This is a tab while the other lines are 4 spaces

extension/js/contentScript.js
```diff
@@ -0,0 +1,20 @@
...
+	 */
+	function injectScript(file_path, tag) {
+		var node = document.getElementsByTagName(tag)[0];
+		var script = document.createElement('script');
+		script.setAttribute('type', 'text/javascript');
+		script.setAttribute('src', file_path);
+		node.appendChild(script);
+		return script;
+	}
+} );
```
MusikAnimal, line 20: The first and last line of this file has MediaWiki-style spacing between parentheses, and the other lines do not. Also I think eslint would want `filePath` instead of `file_path`, etc. I'm sure you know, but can run `./node_modules/.bin/eslint --fix extension/js/contentScript.js` to fix some of these issues automatically.

extension/manifest.json
```diff
@@ -12,8 +12,10 @@
...
   } ],
+  "web_accessible_resources": ["js/generated.pageScript.js"],
```
MusikAnimal: This has no spaces between the brackets, not that it matters much.

build/template_browserextension.js
MusikAnimal: Also, I don't really know what's supposed to happen here, but it would seem we're grabbing every link on the page? Or is this just for the browser extension button?

MusikAnimal (changes requested):
Left a few more comments. One more question... did we want to gitignore extension/js/generated.pageScript.js, which is generated by the build script? Or should it just be committed?

build/template_browserextension.js
mooeypoo: Yeah, I'm an idiot here... I wrote this initially as a placeholder in the task and didn't notice when I copied it over. Ironically, I tested the previous version which was using correct markup. It should have quotes; `$( '<a>' )` creates a <a> element. `$( 'a' )` selects all <a> elements.

Let me redo this part. I'll add a proper temporary behavior until we figure out what kind of button we want here. This is the place where we make a button to turn WhoWroteThat on, so it's a "simple" button placed somewhere in the interface (prateek is working it out still).

extension/js/contentScript.js
mooeypoo: I should add contentScript.js to the tested files in .eslint....

extension/manifest.json
mooeypoo: Yeah it also uses spaces instead of tabs; I think that's the common display for `manifest.json` for browser extension, but I'm not 100% sure. I can fix that.

mooeypoo:
So:

1. I added generated.pageScript.js in; I think that as long as it's clear that it's autogenerated (so on one should manually change it) it's probably okay to keep it in the repo. We can change that later.
2. I added `contentScript.js` to .eslint rules, and fixed to make it pass
3. I made the activation button an actual link somewhere just to test it out and added a 'test.js' file with an `OO.ui.alert` to test the behavior and timing (to make sure ooui is available)
4. I cleaned up `manifest.json` to fit our code styles.
5. I tested the `extension/` directory as an unpacked extension in chrome -- and made sure it works :)

MusikAnimal:
Awesome! This looks good to me, though I haven't tested it.

MusikAnimal (approved):

mooeypoo:
Awesome. I'm merging in. We should take this repo under /Wikimedia (there's a task for that)

PR #4, opened 2019-07-25
 Add CI and fix lint errors in gruntfile and tests 

samwilson:


samwilson:
I've removed the timestamp from the built file so that the build is repeatable (for CI's benefit). I've added in the version number, but we might want to talk about this.

mooeypoo:
Sweet! Thank you!

PR #5, opened 2019-07-26
Fix API methods and set up QUnit tests for CI

samwilson:
This makes the Api class testable (both via QUnit in the browser
and on CLI) and adds tests for turning Wikipedia URLs into
WikiColor API URLs.

https://phabricator.wikimedia.org/T226743

Bug: T226743


mooeypoo:
\o/

PR #8, opened 2019-07-30
Convert to ES6 and add mocha tests and babel to build steps

mooeypoo:
Convert the entire system to use ES6 syntax.

- Use browserify with babel to create the script and inject it into the extension wrapper
- Use mocha and karma for tests
- Convert code to ES6 syntax
- Add relevant build steps that also run the babelification


mooeypoo:
So, I'm actually not 100% convinced with this change. We can do it if we want to: 

- The tests work on pure JS (we'll probably need to get them to work with PhantomJS at some point when we deal with DOM stuff) 
- The output code is babelified and then put into the extension construct (and later in the gadget)
- Writing in ES6/ESNext is pretty cool, and definitely more understandable.

On the other hand, there are a couple of 'meh' things I see in the output:
- The extension file itself looks messy as hell and is hard to examine. This isn't that bad for the browser extension, but is extremely not great for an on-wiki gadget.
- We might have an issue with required libraries; since we do things like `var $ = require( 'jquery' );` for the DOM parts (and OOUI parts) the babelification automatically transcludes the entire jquery library. That's completely unnecessary for both the extension and gadget :\ I was going to take a look into whether we can blacklist transclusions in the babelification process, but I couldn't find anything right now.

So... I'm not sure. What do y'all think? We do get the nice aspect of writing in pure modern javascript, but the price is a bit more complexity in the build, and my biggest worry is that this will "bite us in the butt" in the future in case the complexity (transcluding libraries + unreadable code in gadget) is deemed unacceptable later.... would it?

I'd be a lot less "meh" about this if this was a completely external tool -- but this works inside MediaWiki as an embedded script.

I'm not sure. Opinions required :)

P.S: No idea why this fails CI... the tests pass, the build passes... is this a problem with extension.json? I can't make sense of it. Any ideas?

package.json
```diff
@@ -6,17 +6,32 @@
...
+	"grunt-run": "~0.8.1",
+	"@babel/cli": "^7.1.5",
+    "@babel/core": "^7.1.6",
+    "@babel/node": "^7.0.0",
+    "@babel/preset-env": "^7.1.6",
+    "@babel/register": "^7.0.0",
+	"babelify": "~10.0.0",
+    "chai": "^3.4.1",
+    "mocha": "^2.3.4",
+	"jsdom": "~15.1.1"
```
MusikAnimal: Inconsistent indentation in this file; some are tabs, other spaces

mooeypoo: Yeah, I need to clean up the package.json in general, I think. I was doing a bunch of testing with several packages, some of those aren't actually used in the final result. I'll go over that + the indentations.

MusikAnimal:
> We might have an issue with required libraries; since we do things like var $ = require( 'jquery' ); for the DOM parts (and OOUI parts) the babelification automatically transcludes the entire jquery library.

Why do we need Browserify? I'm probably missing something, but I think we could just use [grunt-babel](https://github.com/babel/grunt-babel) and you won't need to require/export etc., if you say that is causing problems. ES5 works in ES6/7+; you shouldn't have to change any code to make it work with Babel. I envisioned only taking advantage of the syntactical candy like arrow functions and argument unpacking.

> The extension file itself looks messy as hell and is hard to examine. This isn't that bad for the browser extension, but is extremely not great for an on-wiki gadget.

A lot of the weirdness appears to be because we're using a `class` and `require`, which I don't think we have to use. Even if we didn't, you're right there will be some possibly incomprehensible stuff in there. A comment at the top with a link to the source would help.

> No idea why this fails CI

When I run `npm install` on my local (Node 11.5.0), I see uncommitted changes to package.json and package-lock.json. I think the build is failing for the same reason.

----

For the record, I've no strong feelings about using Babel, I just saw this as an opportunity to use modern JS since we're able to. I did not envision it being this involved!

mooeypoo:
That's a good point; when you were talking about ES6/7 I thought about the beauty of classes and easy syntax and construction that we get with 'import'. We could do "half" way, where we still use ES5 but with some es6/7 syntax... I can try and see what that means.

mooeypoo:
OK, This is ready for re-review!

samwilson (approved):
Looks great to me! I think ES6 is going to be a good decision.

This removes the default wikiwho URL config; I'm sure we'll re-add it in a subsequent patch though.

The only thing I can think to improve is that this should be a single commit, with a more detailed message (i.e. just what's in the PR description).

src/app.js
```diff
@@ -0,0 +1,6 @@
+import Api from './Api';
+
+// eslint-disable-next-line no-unused-vars
+const a = new Api(); // Test
```
samwilson: We were creating this object with `{ url: extwrt.globals.wikicolorUrl }`, but that global config is gone now. Maybe just pass the URL in manually here for now? I'm not sure.

mooeypoo: Oops, I think this went away in my testing and then the rebase. I'm bringing that back in.

mooeypoo:
OK:
- I squashed the commits into one
- added the information into the commit message
- Added a `src/config.js` file for the config options, with the wikiwho URL in there

This is ready again :)

PR #9, opened 2019-07-31
Add JSDoc grunt task

samwilson:
This is just an experiment in JSDoc. I just wanted to see how its docblock validation was. (Answer: not amazing but somewhat useful.)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -9,6 +9,7 @@ module.exports = function Gruntfile( grunt ) {
 	grunt.loadNpmTasks( 'grunt-concat-with-template' );
 	grunt.loadNpmTasks( 'grunt-contrib-less' );
 	grunt.loadNpmTasks( 'grunt-stylelint' );
+	grunt.loadNpmTasks( 'grunt-jsdoc' );
 
 	grunt.initConfig( {
 		pkg: pkg,
@@ -18,6 +19,7 @@ module.exports = function Gruntfile( grunt ) {
 					'**/*.js',
 					'!build/**',
 					'!extension/**',
+					'!docs/**',
 					'!node_modules/**',
 					'!temp/**',
 					'extension/js/contentScript.js'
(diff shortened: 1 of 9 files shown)


Gruntfile.js
```diff
@@ -31,6 +33,21 @@ module.exports = function Gruntfile( grunt ) {
...
+				src: [ 'src/*.js', 'test/*.js' ],
+				options: {
+					encoding: 'utf8',
+					readme: 'README.md',
+					tutorials: 'help',
+					recurse: true,
+					template: 'node_modules/jsdoc-wmf-theme',
+					destination: 'doc',
+					'package': 'package.json',
+					ignoreWarnings: false
```
prtksxna: Might want to add the markdown and plugin? Maybe the hierarchy one if this is going to get complicated 😜

https://github.com/wikimedia/oojs/blob/master/.jsdoc.json#L14

samwilson: Done.

tutorials/install.md
```diff
@@ -0,0 +1,15 @@
+Installing WhoWroteThat
+=======================
+
+## As a gadget
+
+1. Copy the contents of [`gadget-WhoWroteThat.js`](../build) to your wiki.
```
mooeypoo: Not quite; we're not outputting the gadget to `/build` (we're actually not outputting anywhere for the moment, we need to add that build step specifically) -- the `/build` folder is for build accessories, like the header-banners and templates.

I think we shoul output the gadget to `/gadget` (since the extension is at `extension/`) *or* put both folders in `/dist` and have `dist/extension` and `dist/gadget`

I set that up in #13

mooeypoo (changes requested):
Looks good, except for a tiny change in the "install" instructions.

tutorials/install.md
samwilson: I actually meant to put a full path in there, but forgot. Fixed now (also with the CSS file); see what you think.

mooeypoo (approved):
Looks good to me!

cimurah (approved):
I checked out this to my local environment and it looks good.

PR #13, opened 2019-08-01
BUILD: Output gadget files

mooeypoo:
*NOTE*: This is written on top of #8 and should only be merged after that PR is merged

- Add a build step to output gadget files on top of browser extension files.
- Move generated output into `dist/` folder for cleanliness


build/header.gadget.txt
```diff
@@ -0,0 +1,18 @@
...
+ * https://github.com/wikimedia/WhoWroteThat
+ *
+ * ***************************************************
+ * ** Please do not change this local file directly **
+ * ***************************************************
+ *
+ * Any updates should be created from the official repository
+ * by running `grunt build` and using
+ * `dist/gadget/generated.WhoWroteThat.gadget.js` and
+ * `dist/gadget/generated.WhoWroteThat.gadget.css` and
```
samwilson: These paths are different to what's elsewhere in this patch, I think.

mooeypoo: Whoops, that's right, I'm fixing.

mooeypoo:
Comments fixed, I squashed the commits into one with a more comprehensive commit message (copied into the PR)

Ready for review; ONLY MERGE THIS AFTER THE ES6 PR! :)

Gruntfile.js
```diff
@@ -49,67 +77,49 @@ module.exports = function Gruntfile( grunt ) {
...
-							src = src
-								.replace( /^\( function \(\) {/, '' ) // Beginning of file
-								.replace( /}\(\) \);\n$/, '' );
-						}
-						// eslint-disable-next-line quotes
-						return '/* >> Starting source: ' + filepath + " << */\n" +
-							src +
-							'/* >> End source: ' + filepath + ' << */';
-					}
+					banner: grunt.file.read( 'build/header.gadget.txt' )
```
samwilson, line 85: Should this header also be added to the generated JS file?

samwilson (approved):
Works in my testing! :)

Gruntfile.js
mooeypoo: So, I thought about that, but the generated JS is inserted *into* the template.

I was thinking of putting it up at the top of the template, but I need to look into how to make it replace the variables if I do that, so I figured we can do that as maybe a followup.

samwilson: Sure, let's leave it for after. This is all working in the important ways.

PR #15, opened 2019-08-02
BUILD: Add i18n file support

mooeypoo:
- Add `i18n` folder with translation files and qqq.json
- Add a function to Gruntfile to create a json object of all available
  language files and translations.
- Replace concat_with_template with `replace` job to handle the template
  - Replace the script with the generated script from browserify
  - Replace the language blob with the generated language json
- Load the current interface language (with English fallback) into
  `mw.messages.set`


samwilson (approved):
Looks good to me, and works. I tested by including a mw.message() call successfully. Not sure if this patch is meant to use the `ext-whowrotethat-activation-link` message for the link, but that will come later anyway.

The rigmarole around getting replace to run after browserify feels a bit complex, but that's just the joy of Grunt I guess. :)

PR #16, opened 2019-08-04
Add activation button routine

mooeypoo:
**Depends on #15** Only merge this PR after #15 

- Create a singleton to handle the activation button
- Only add the button to main namespace in view mode
- Inject the button to the page with `mw.util.addPortletLink`

Task: https://phabricator.wikimedia.org/T226761

### DIFF
--- a/src/singleton.activation.js
+++ b/src/singleton.activation.js
@@ -1,0 +1,69 @@
+// eslint-disable-next-line no-unused-vars
+var wwtActivationSingleton = ( function () {
+	var originalHTML = $( '#bodyContent.mw-body-content' ).clone(),
+		interfaceLang = $( 'html' ).attr( 'lang' ),
+		/**
+		 * Initialize the activation button to toggle WhoWroteThat system.
+		 * Only attaches the button for pages in the main namespace, and only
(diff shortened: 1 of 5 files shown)


src/singleton.activation.js
```diff
@@ -0,0 +1,69 @@
...
+		 * Get the original html of the article, for the purposes of toggling
+		 * the system on and off.
+		 *
+		 * @return {jQuery} Content node
+		 */
+		getOriginalHTML: function () {
+			return originalHTML;
+		},
+		initialize: function ( translations, onClickFunction ) {
+			// Bail out if we're anywhere that is not an article page in read mode
```
samwilson, line 53: I'm not sure what we're meant to do on diff pages, but this conditional won't stop the WWT link being displayed on diff pages where there's no article body (i.e with the `diffonly` preference enabled). Maybe we could just search for a `#mw-content-text .mw-parser-output` element as that's what we replace with the returned HTML.

mooeypoo: Diff pages are also "view" action?? gah. Damn. I guess if that's the case, I'll look for the css classes :\

PR #18, opened 2019-08-07
Add activation button process

samwilson:
- Create a singleton to handle the activation button
- Only add the button to main namespace in view mode, and not
  on the main page
- Inject the button to the page with `mw.util.addPortletLink`

### DIFF
--- a/src/singleton.activation.js
+++ b/src/singleton.activation.js
@@ -1,0 +1,68 @@
+// eslint-disable-next-line no-unused-vars
+var wwtActivationSingleton = ( function () {
+	var $parserOutput = $( '.mw-parser-output' ),
+		$parserOutputStored = $parserOutput.clone(),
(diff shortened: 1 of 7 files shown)


src/singleton.activation.js
```diff
@@ -0,0 +1,70 @@
...
+
+			// Add a portlet link to 'tools'
+			link = mw.util.addPortletLink(
+				'p-tb',
+				'#',
+				mw.msg( 'ext-whowrotethat-activation-link' ),
+				't-whowrotethat',
+				mw.msg( 'ext-whowrotethat-activation-link-tooltip' ),
+				'',
+				'#t-print'
```
samwilson: I don't think it really matters, but in English Wikipedia the t-print link has been moved out of the toolbox and into a 'Print/export' portlet. So our link is added at the end (with whatever other gadgets do the same).

PR #19, opened 2019-08-08
Add InfoBarWidget

mooeypoo:
- Add a widget for the info bar
- Set the widget with 3 states: 'ready', 'err', and 'pending',
  according to the given styles.
- Make the widget sticky when the window is scrolled down
- When the [x] is clicked, hide the widget and reload original
  content.

### DIFF
--- a/src/InfoBarWidget.js
+++ b/src/InfoBarWidget.js
@@ -1,0 +1,103 @@
+var InfoBarWidget = function InfoBarWidget( config = {} ) {
+	// Parent constructor
+	OO.ui.ButtonWidget.parent.call( this, config );
+	OO.ui.mixin.IconElement.call( this, config );
+	OO.ui.mixin.LabelElement.call( this, config );
+	OO.ui.mixin.TitledElement.call( this, config );
+	OO.ui.mixin.FlaggedElement.call( this, config );
+
+	this.closeIcon = new OO.ui.IconWidget( {
+		icon: 'clear',
+		flags: [ 'invert' ],
+		classes: [ 'ext-wwt-infoBarWidget-close' ]
+	} );
+	this.userInfoLabel = new OO.ui.LabelWidget( {
+		label: mw.msg( 'ext-whowrotethat-ready-general' ),
+		classes: [ 'ext-wwt-infoBarWidget-info' ]
+	} );
+	this.$pendingAnimation = $( '<div>' )
+		.addClass( 'ext-wwt-infoBarWidget-spinner' )
+		.append(
+			$( '<div>' )
+				.addClass( 'ext-wwt-infoBarWidget-spinner-bounce' )
+		);
+
+	// Set properties
+	this.setState( config.state || 'pending' );
+	this.setLabel( $( '<span>' ).append( mw.msg( 'ext-whowrotethat-state-pending' ) ).contents() );
+
+	// Close event
+	this.closeIcon.$element.on( 'click', () => this.emit( 'close' ) );
+
+	// Initialize
+	this.$element
+		.addClass( 'ext-wwt-infoBarWidget' )
+		.append(
+			this.$pendingAnimation,
+			this.$icon,
+			this.$label,
+			this.userInfoLabel.$element,
+			this.closeIcon.$element
+		);
+};
+
+/* Setup */
+
+OO.inheritClass( InfoBarWidget, OO.ui.Widget );
+OO.mixinClass( InfoBarWidget, OO.ui.mixin.IconElement );
+OO.mixinClass( InfoBarWidget, OO.ui.mixin.LabelElement );
+OO.mixinClass( InfoBarWidget, OO.ui.mixin.TitledElement );
+OO.mixinClass( InfoBarWidget, OO.ui.mixin.FlaggedElement );
+
+/**
+ * Define legal states for the widget
+ *
+ * @type {Array}
+ */
+InfoBarWidget.static.legalFlags = [ 'pending', 'ready', 'err' ];
+
+/**
+ * Change the state of the widget
+ *
+ * @param {string} state Widget state; 'pending', 'ready' or 'error'
+ */
+InfoBarWidget.prototype.setState = function ( state ) {
+	const flags = {};
+
+	if ( this.state !== state ) {
+		this.constructor.static.legalFlags.forEach( flag => {
+			flags[ flag ] = flag === state;
+		} );
+		flags.invert = true;
+
+		this.setFlags( flags );
+
+		if ( state === 'ready' ) {
+			this.setLabel( $( '<span>' ).append( mw.msg( 'ext-whowrotethat-ready-title' ) ).contents() );
+			this.userInfoLabel.setLabel( $( '<span>' ).append( mw.msg( 'ext-whowrotethat-ready-general' ) ).contents() );
+			this.setIcon( 'userAvatar' );
+		} else if ( state === 'pending' ) {
+			this.setIcon( '' );
+		} else {
+			this.setIcon( 'error' );
+			this.setErrorMessage();
+		}
+
+		this.$pendingAnimation.toggle( state === 'pending' );
+		this.userInfoLabel.toggle( state === 'ready' );
+		this.closeIcon.toggle( state !== 'pending' );
+
+		this.state = state;
+	}
+};
+
+/**
+ * Set an error with a specific label
+ *
+ * @param {string} message Error label message key
+ */
+InfoBarWidget.prototype.setErrorMessage = function ( message = 'ext-whowrotethat-state-error-generic' ) {
+	this.setLabel( mw.msg( 'ext-whowrotethat-state-error', mw.msg( message ) ) );
+};
+
+export default InfoBarWidget;
--- a/src/app.js
+++ b/src/app.js
@@ -1,8 +1,33 @@
 import config from './config';
 import Api from './Api';
+import InfoBarWidget from './InfoBarWidget';
+let components = $( 'body' ).data( 'wwt-components' );
 
-// eslint-disable-next-line no-unused-vars
-const a = new Api( { url: config.wikicolorUrl } ); // Test
-// TEST!
+// Initialize
+if ( !components ) {
+	// Store, so we don't append and create objects if they've been created
+	// TODO: We could potentially store all that data in some view model that's shared
+	// across the widgets.
+	components = {
+		widget: new InfoBarWidget(),
+		api: new Api( { url: config.wikicolorUrl } ),
+		$originalOutput: $( 'body' ).data( 'wwt-originalOutput' )
+	};
 
-OO.ui.alert( 'The extension is working! URL: ' + a.getAjaxURL( window.location.href ) );
+	$( 'body' ).data( 'wwt-components', components );
+
+	if ( $( 'body' ).hasClass( 'skin-timeless' ) ) {
+		$( '#mw-content-wrapper' ).prepend( components.widget.$element );
+	} else {
+		$( '#content' ).prepend( components.widget.$element );
+	}
+}
+
+components.widget.toggle( true );
+components.widget.setState( 'pending' );
+components.widget.on( 'close', () => {
+	// Close button; revert back to the original content
+	// $( '.mw-parser-output' ).empty().append( components.$originalOutput );
+	// Hide the widget
+	components.widget.toggle( false );
+} );
--- a/src/less/InfoBarWidget.less
+++ b/src/less/InfoBarWidget.less
@@ -1,0 +1,142 @@
+@spinner-size: 12px;
+
+.ext-wwt-infoBarWidget {
+	position: sticky;
+	top:0;
+	display: flex;
+	z-index: 1;
+	width: 100%;
+	box-shadow: 0 2px 6px 0 rgba(0, 0, 0, 0.25);
+	box-sizing: border-box;
+	padding: 16px;
+	color: #ffffff;
+	.skin-vector & {
+		// This duplicates Vector, making sure that the
+		// bar stops at the edge of the content even though
+		// it is outside the content box itself.
+
+		width: calc( 100% + 1.5em + 1.5em );
+		// content body has 1.25em padding top
+		margin-top: -1.25em;
+		// put in some space before the heading
+		margin-bottom: 1em;
+
+		.ltr& {
+			// content body has 1.5em padding left and right
+			margin-left: -1.5em;
+		}
+		.rtl& {
+			// content body has 1.5em padding left and right
+			margin-right: -1.5em;
+		}
+	}
+
+	.skin-monobook & {
+		margin-left: -1em;
+		margin-bottom: 2em;
+		// Correct for the negative margin + an extra 1em padding on the content div
+		width: calc( 100% + 1em + 1em );
+
+		// Since monobook uses 'Apex', we don't have the white (inverted)
+		// version of the 'close' icon. For that skin, we'll inject data-uri
+		.ext-wwt-infoBarWidget-close.oo-ui-iconElement {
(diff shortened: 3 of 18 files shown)


prtksxna:
I checked it out in the `pending`, `ready` and `err` states and this is looking good 👍🏽

I did notice a [bit of a jump on Chrome](https://drive.google.com/file/d/18Pgg9NczKBUnTbzetDUeSSB9Mtjv-8Ct/view?usp=sharing) in the sticky behavior. A `position: sticky;` might work for this, but I know you've been battling Vector and Monobook 😓😛

src/InfoBarWidget.js
```diff
@@ -0,0 +1,122 @@
+var InfoBarWidget = function InfoBarWidget( config ) {
+	var widget = this;
```
MusikAnimal: do we have ES6+ here? Because we should be using `let` or `const` (`const` in both lines above, I think)

src/InfoBarWidget.js
```diff
@@ -0,0 +1,122 @@
...
+	}
+};
+
+/**
+ * Set an error with a specific label
+ *
+ * @param {string} message Error label message key
+ */
+InfoBarWidget.prototype.setErrorMessage = function ( message ) {
+	message = message || 'ext-whowrotethat-state-error-generic';
```
MusikAnimal: you should be able to set defaults in the argument list, `function ( message = 'ext-whowrotethat-state-error-generic' )`

src/InfoBarWidget.js
```diff
@@ -0,0 +1,122 @@
+var InfoBarWidget = function InfoBarWidget( config ) {
+	var widget = this;
+
+	// Configuration initialization
+	config = config || {};
```
MusikAnimal: you could instead change line 1 to `function InfoBarWidget( config = {} ) {`

src/InfoBarWidget.js
```diff
@@ -0,0 +1,122 @@
...
+	this.setLabel( mw.msg( 'ext-whowrotethat-state-pending' ) );
+
+	// Scroll event
+	$( window ).scroll( () => {
+		var pos = $( '#bodyContent' )[ 0 ].getBoundingClientRect();
+
+		widget.$element.toggleClass(
+			'ext-wwt-infoBarWidget-sticky',
+			pos.top < widget.$element.height()
+		);
```
MusikAnimal: it seems to run fast, but maybe we shouldn't call `.toggleClass` on every pixel that is scrolled. So something like
```
if ( pos.top < widget.$element.height() ) {
	widget.$element.addClass( 'ext-wwt-infoBarWidget-sticky' );
} else {
	widget.$element.removeClass( 'ext-wwt-infoBarWidget-sticky' );
}
```
or maybe jQuery is smart enough to not re-manipulate the DOM (assuming this is jQuery)

src/InfoBarWidget.js
```diff
@@ -0,0 +1,122 @@
...
+	} );
+
+	// Close event
+	this.closeIcon.$element.on( 'click', function () {
+		widget.emit( 'close' );
+	} );
```
MusikAnimal: not really important but you could do `this.closeIcon.$element.on( 'click', () => widget.emit( 'close' ) )`. Perhaps that's too ugly.

Also if you changed this to an arrow function I don't think you'd need `widget`, since here and in other places you use `widget`, it is the same as `this`.

src/InfoBarWidget.js
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Change the state of the widget
+ *
+ * @param {string} state Widget state; 'pending', 'ready' or 'error'
+ */
+InfoBarWidget.prototype.setState = function ( state ) {
+	var flags = {};
+
+	if ( this.state !== state ) {
+		this.constructor.static.legalFlags.forEach( ( flag ) => {
```
MusikAnimal: `( flag )` can just be `flag`, if it matters (maybe we should set some additional rules in eslint?)

src/InfoBarWidget.js
MusikAnimal: Or maybe we don't need `widget` at all, see below

src/InfoBarWidget.js
```diff
@@ -0,0 +1,122 @@
...
+
+	// Configuration initialization
+	config = config || {};
+
+	// Parent constructor
+	OO.ui.ButtonWidget.parent.call( this, config );
+	OO.ui.mixin.IconElement.call( this, config );
+	OO.ui.mixin.LabelElement.call( this, config );
+	OO.ui.mixin.TitledElement.call( this, config );
+	OO.ui.mixin.FlaggedElement.call( this, config );
```
MusikAnimal, line 7: couldn't you just do `OO.ui.mixin.IconElement( config )` (same for the other lines)?

src/app.js
```diff
@@ -1,8 +1,34 @@
...
+	components = {
+		widget: new InfoBarWidget(),
+		api: new Api( { url: config.wikicolorUrl } ),
+		$originalOutput: $( 'body' ).data( 'wwt-originalOutput' )
+	};
 
-OO.ui.alert( 'The extension is working! URL: ' + a.getAjaxURL( window.location.href ) );
+	$( 'body' ).data( 'wwt-components', components );
+
+	if ( $( 'body' ).hasClass( 'skin-vector' ) ) {
```
MusikAnimal: `$( '#mw-head-base' ).after( components.widget.$element )` should work in all skins. If not, could I beg to also put `|| $( 'body' ).hasClass( 'timeless' )`? :)

src/app.js
```diff
@@ -1,8 +1,34 @@
...
+		components.widget.$element
+			.insertBefore( $( '#content.mw-body' ) );
+	} else if ( $( 'body' ).hasClass( 'skin-monobook' ) ) {
+		$( '#content.mw-body' ).prepend( components.widget.$element );
```
MusikAnimal: There's only one `#content` so the `.mw-body` is redundant (same with line 21)

src/less/InfoBarWidget.less
```diff
@@ -0,0 +1,149 @@
...
+				border-radius: 100%;
+				.animation( wwtBouncedelay 1600ms ease-in-out -160ms infinite both );
+			}
+
+			&:before {
+				margin-right: 4px;
+				.animation-delay( -330ms );
+			}
+
+			&:after {
```
MusikAnimal: Pseudo-elements should have two colons, e.g. `&::after` https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#Syntax (same with lines 62-63, 75)

src/less/InfoBarWidget.less
```diff
@@ -0,0 +1,149 @@
...
+	50%,
+	100% {
+		-moz-transform: scale( 0.625 );
+	}
+
+	20% {
+		opacity: 0.87;
+		-moz-transform: scale( 1 );
+	}
+}
```
MusikAnimal: `@keyframes` is supported in all modern browsers, even IE 10, so we shouldn't need the `-webkit` and `-moz` prefixes. Same for `transform`

MusikAnimal (changes requested):
Looking good! I did notice the same jumpiness that Prateek pointed out.

Just a few minor nitpicks, which I apologize for! Just trying to take advantage of ES6+

src/less/InfoBarWidget.less
mooeypoo: Ooh, interesting

```
Note: As a rule, double colons (::) should be used instead of a single colon (:). This distinguishes pseudo-classes from pseudo-elements. However, since this distinction was not present in older versions of the W3C spec, most browsers support both syntaxes for the original pseudo-elements.
```
That explains why I saw it either way, and both ways worked. I'll change it.

src/app.js
mooeypoo: Good point, though I wanted to make sure I'm being contextually specific. I can remove it, or keep, either way works, it's more for context than function.

src/app.js
mooeypoo: Unfortunately, I just checked in the console:

`$( '#mw-head-base' )` exists in Vector.
`$( '#mw-head-base' )` does not exist in Timeless.
`$( '#mw-head-base' )` does not exist in Monobook.

I'll put in a correction for Timeless, and even if we find a consistent placement, we will still need some corrections per skin, probably, because the widths of the elements are not consistent, and the bar needs to fit the width of the content (and the head is outside it)

With my testing, it seemed most straight forward (with the least per-skin CSS corrections) to find a different placement for each of those. We can iterate later, but I'll at least make sure that in this patch it works without being too mangled.

src/InfoBarWidget.js
mooeypoo: Unfortunately no, because it's not exactly a function, it's an object constructor. If you call it directly, its context would be itself, and we need the context to be this object; that's why we use `.call` and give it the current context (`this`).

mooeypoo: You can try this in the console:

```
var ContextTester = function ( name ) { this.name = name; };
ContextTester.prototype.printName = function () { console.log( this.name ); }

var Tester = function () { this.name = 'tester'; }
Tester.prototype.test = function () {
   var o = new ContextTester( 'one' );
   o.printName(); // 'one'
   o.printName.call( this ); // 'tester'
};

var t = new Tester();
t.test();
```

src/InfoBarWidget.js
mooeypoo: Aye, it probably can be; I was just really used to the parentheses. I'm actually not sure what our eslint rules are on that; I'll give it a try and see if eslint lets it in.

src/InfoBarWidget.js
mooeypoo: Good point, I forgot I can do that

src/app.js
MusikAnimal: I could have sworn I tested this! Apparently not. We could put the bar before `#firstHeading` but this would compromise the design due to the padding around the outermost `.mw-body`.

src/InfoBarWidget.js
MusikAnimal: I don't think we're using any for this, but we could https://eslint.org/docs/rules/arrow-parens. Some enforce the parentheses others enforce it not be there. Very trivial, I just love how there's a lot less syntax in ES6+!

src/InfoBarWidget.js
MusikAnimal: Ah, that makes sense. Thanks for the explanation!

src/app.js
MusikAnimal: It doesn't matter :) I think using only the ID will be faster (jQuery should use `getElementById`), but that has got to be by only microseconds.

src/InfoBarWidget.js
mooeypoo: Yup, I'm using `position: sticky` instead!

src/InfoBarWidget.js
mooeypoo: I added the rule to the `.eslintrc` file:

```
        "arrow-parens": [ "error", "as-needed" ]
```

src/app.js
mooeypoo: Yeah this isn't awesome :\

Okay, I went over all three skins and adjusted. I had to attach the widget a bit differently in each one because of visual considerations. In Timeless, for example, you already have a floating 'sticky' header, so the widget needed to be below it but constrained to the content area. In monobook, the spacing was a bit off, so I corrected with negative margins.

I also had to pull in the representation of the "close" icon into data-uri for Timeless and Monobook; it looks like Apex (ooui theme used in those two skins) doesn't have that icon. Bah.

src/app.js
mooeypoo: Changed.

src/less/InfoBarWidget.less
mooeypoo: Good point. We're still using that in core, so a part of me wonders why... but in this case, it's probably safe anyways. Removing that.

mooeypoo:
EUGHHHH I just discovered a disgustingly stupid bug.

I'm attaching the widget to the body (by inserting "before" the content element) and making it "sticky". I just tested it with a very long article -- and midway through, the sticky part stopped working and the widget vanished.

After investigation, it seems that the `#content` div continues after the `<body>` ends. I am not sure how that happens, I thought 'body' should go all the way with its content, but apparently not. So, the widget uses the 'body' as its container and once it "ends" it stops.

The solution (which I will implement in the next correction patch) is this:

1. Prepend the widget to `#content.mw-body` (rather than inserting before)
2. It will be impacted by the padding of the body, so correct those by using the following styles (tested those in the console):

```
.ltr.skin-vector .ext-wwt-infoBarWidget {
    // content body has 1.5em padding left and right
    margin-left: -1.5em;
    width: calc( 100% + 1.5em + 1.5em );
    // content body has 1.25em padding top
    margin-top: -1.25em;
    // put in some space before the heading
    margin-bottom: 1em;
}
```

mooeypoo:
Fixed that bug. Monobook is annoying, but it works now.

src/less/mixins.less
```diff
@@ -0,0 +1,19 @@
...
+	-moz-transform: rotate( @deg );
+	transform: rotate( @deg );
+}
+
+// @see https://github.com/wikimedia/mediawiki/blob/37e1fffd827b483f805c6c5296155768a74a2bc6/resources/src/mediawiki.rcfilters/styles/mw.rcfilters.mixins.less
+.animation-delay( ... ) {
+	-webkit-animation-delay: @arguments; // Chrome 4-42, Safari 4-8, Opera 15-29, Android 2.1-4.4.4
+	-moz-animation-delay: @arguments; // Firefox 5-15
+	animation-delay: @arguments; // Chrome 43+, Firefox 16+, IE 10+, Edge 12+, Safari 9+, Opera 12.10 & 30+, iOS 9+, Android 47+
+}
```
MusikAnimal: We shouldn't need these browser prefixes, hence I don't think we need these mixins, right?

src/less/InfoBarWidget.less
```diff
@@ -0,0 +1,145 @@
+@import 'mixins';
+
+@spinner-size: 12px;
+
+.ext-wwt-infoBarWidget {
+	position:-webkit-sticky;
+	position: sticky;
```
MusikAnimal, line 4: It may be worth noting this isn't supported in IE. I for one could care less :stuck_out_tongue_closed_eyes: `position: sticky` looks so buttery smooth compared to JS implementations, so I think we should keep it.

I don't think we need the `-webkit-sticky` though, since this isn't needed as of Safari 7.1

MusikAnimal (changes requested):
The font and icon sizes look a little off in Monobook, but I understand if we don't care:

Screenshot from 2019-08-13 13-32-24

Meanwhile it looks uh-mazing in Vector and Timeless!

I'm a little nervous about negative margins and such, trying to adapt to the styles of the skin, since these things can change, but what we have here seems more than fine for now.

One more tiny nitpick about browser prefixes :)

mooeypoo:
> The font and icon sizes look a little off in Monobook, but I understand if we don't care:
> 
> Screenshot from 2019-08-13 13-32-24

Yeah, Monobook doesn't have the "invert" icons, so they don't appear while as is. I corrected that for the [X] icon, but the person-thing icon is not the one we need to use anyways (I need the icon again from @prtksxna, and planned to change that in a followup) so I didn't bother too much with fixing monobook with it right now, as it will be changed and then corrected in monobook in a followup.

> I'm a little nervous about negative margins and such, trying to adapt to the styles of the skin, since these things can change, but what we have here seems more than fine for now.

I don't know if we have much choice, honestly :\ placing it in a different DOM wrapper would mean putting on the same margin constraints (that cna change) on a positive, *and* it has some drawbacks, so I think it's as safe as it can get as-is. Those margins rarely change anyways.

> One more tiny nitpick about browser prefixes :)

That one didn't work for me without the prefixes, so I'm going to be careful and leave them in for now at least :\

src/less/InfoBarWidget.less
mooeypoo: It didn't work for me in chrome without it. It might have had a separate issue, but I wanted to be a bit more extra careful here. 

I'll test it again, see if it was something else that didn't work.

src/less/mixins.less
mooeypoo: Eh. I have to say, those cost us nothing, and they're used all over MediaWiki. I'm a bit weary of taking them all out...? Are we 100% sure they work without the prefixes?

MusikAnimal (approved):

PR #25, opened 2019-08-13
ES6-ify more code; change 'Wikicolor' to 'WikiWho'

MusikAnimal:


MusikAnimal:
Hope this is okay! Mostly unnecessary stuff, just keeping coding styles consistent.

I noticed singleton.activation.js doesn't get Babelified. If that can't be fixed, maybe it should live outside the src/ folder?

mooeypoo:
> I noticed singleton.activation.js doesn't get Babelified. If that can't be fixed, maybe it should live outside the src/ folder?

Yup I was planning to clean that up, make it proper ES6, and babel it. It will need to be outputted outside of the others, so maybe I should put it in a subfolder in `src/` but it should definitely be fixed up.

mooeypoo (approved):

PR #26, opened 2019-08-14
Add error and result handling for API response

samwilson:
Along with some example code in app.js that uses the new API
result structure.

https://phabricator.wikimedia.org/T226743

Bug: T226743

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -8,5 +8,6 @@
 	"ext-whowrotethat-state-error": "Error: $1",
 	"ext-whowrotethat-state-error-generic": "Could not load WhoWroteThat. Please refresh and try again.",
 	"ext-whowrotethat-ready-title": "Who Wrote That?",
-	"ext-whowrotethat-ready-general": "Hover to see contributions by the same author. Click for more details."
+	"ext-whowrotethat-ready-general": "Hover to see contributions by the same author. Click for more details.",
+	"ext-whowrotethat-api-error-wikiwho-generic": "WikiWho data is not currently available for this article. Please try again soon."
 }
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -8,5 +8,6 @@
 	"ext-whowrotethat-state-error": "Displays in the information bar in case there were errors while loading.",
 	"ext-whowrotethat-state-error-generic": "A generic error suggesting to the user that they should refresh.",
 	"ext-whowrotethat-ready-title": "Title of the extension that appears when the system is loaded and displays information about the selection.",
-	"ext-whowrotethat-ready-general": "Instruction label that appears when the system is ready, before any hover or selection was made yet."
+	"ext-whowrotethat-ready-general": "Instruction label that appears when the system is ready, before any hover or selection was made yet.",
+	"ext-whowrotethat-api-error-wikiwho-generic": "Error message displayed when an article's data could not be fetched."
 }
--- a/src/Api.js
+++ b/src/Api.js
@@ -16,6 +16,8 @@ class Api {
 		if ( this.url && this.url.slice( -1 ) === '/' ) {
 			this.url = this.url.slice( 0, -1 );
 		}
+
+		this.results = null;
 	}
 
 	/**
@@ -89,13 +91,102 @@ class Api {
 	}
 
 	/**
-	 * Get the WhoColor data for a given wiki page.
+	 * Extract token and editor IDs from a WikiWho span element with `id='token-X'` and
+	 * `class='token-editor-Y'` attributes.
 	 *
-	 * @param  {string} wikiUrl URL of the wiki page that we want to analyze.
-	 * @return {Promise}
+	 * @param {HTMLElement} element
+	 * @return {Object} An object with two parameters: tokenId and editorId (string).
+	 */
+	getIdsFromElement( element ) {
+		const out = { tokenId: false, editorId: false },
+			tokenMatches = element.id.match( /token-(\d+)/ ),
+			editorMatches = element.className.match( /token-editor-([^\s]+)/ );
+		if ( tokenMatches && tokenMatches[ 1 ] ) {
+			out.tokenId = parseInt( tokenMatches[ 1 ] );
+		}
+		if ( editorMatches && editorMatches[ 1 ] ) {
+			out.editorId = editorMatches[ 1 ];
+		}
+		return out;
+	}
+
+	/**
+	 * Get the WikiWho replacement for `.mw-parser-output` HTML.
+	 * @return {string}
+	 */
+	getReplacementHtml() {
+		return this.results.extended_html;
+	}
+
+	/**
+	 * Get user and revision information for a given token.
+	 *
+	 * @param {int} tokenId
+	 * @return {{revisionId: *, score: *, userId: *, username: *, revisionTime: *}|boolean} Object
+	 * that represents the token info or false if a token wasn't found.
+	 */
+	getTokenInfo( tokenId ) {
+		let revId, revision, username, score;
+
+		// Get the token information. results.tokens structure:
+		// [ [ conflict_score, str, o_rev_id, in, out, editor/class_name, age ], ... ]
+		// e.g. Array(7) [ 0, "indicate", 769691068, [], [], "18201938", 76652371.587203 ]
+		const token = this.results.tokens[ tokenId ];
+		if ( !token ) {
+			return false;
+		}
+
+		// Get revision information. results.revisions structure:
+		// { rev_id: [ timestamp, parent_rev_id, user_id, editor_name ], ... }
+		// e.g. Array(4) [ "2017-03-11T02:12:47Z", 769315355, "18201938", "Biogeographist" ]
+		revId = token[ 2 ];
+		revision = this.results.revisions[ revId ];
+		username = revision[ 3 ];
+
+		// Get the user's edit score (percentage of content edited).
+		// results.present_editors structure:
+		// [ [ username, user_id, score ], ... ]
+		for ( let i = 0; i < this.results.present_editors.length; i++ ) {
+			if ( this.results.present_editors[ i ][ 0 ] === username ) {
+				score = parseFloat( this.results.present_editors[ i ][ 2 ] ).toFixed( 1 );
+				break;
+			}
+		}
+
+		// Put it all together.
+		return {
+			username: username,
+			userId: token[ 5 ],
+			revisionId: revId,
+			revisionTime: new Date( revision[ 0 ] ),
+			score: score
+		};
+	}
+
+	/**
+	 * Get the WikiWho data for a given wiki page.
+	 *
+	 * @param {string} wikiUrl URL of the wiki page that we want to analyze.
+	 * @return {Promise} A promise that resolves when the data is ready,
+	 * or rejects if there was an error.
 	 */
 	getData( wikiUrl ) {
-		return $.getJSON( this.getAjaxURL( wikiUrl ) );
+		const api = this;
+		if ( this.resultsPromise ) {
+			return this.resultsPromise;
+		}
+		this.resultsPromise = $.getJSON( this.getAjaxURL( wikiUrl ) )
+			.then( function ( result ) {
+				// Handle error response.
+				if ( !result.success && result.info ) {
+					// The API gives us an error message, but we don't use it because it's only
+					// in English.
+					return $.Deferred().reject( 'ext-whowrotethat-api-error-wikiwho-generic' );
+				}
+				// Store all results.
+				api.results = result;
+			} );
+		return this.resultsPromise;
 	}
 }
 
--- a/src/app.js
+++ b/src/app.js
@@ -25,6 +25,36 @@ if ( !components ) {
 
 components.widget.toggle( true );
 components.widget.setState( 'pending' );
+components.api.getData( window.location.href )
+	.then(
+		// Success handler.
+		function () {
+			const api = components.api;
+			// Insert modified HTML.
+			$( '.mw-parser-output' ).html( api.getReplacementHtml() );
+			// Highlight when hover a user's contributions.
+			// @TODO This is just testing code and should be replaced by the proper behaviour.
+			$( '.mw-parser-output .editor-token' )
+				.on( 'mouseenter', function () {
+					const ids = api.getIdsFromElement( this );
+					// Activate all this user's contribution spans.
+					$( '.token-editor-' + ids.editorId ).addClass( 'active' );
+					// Information for popup.
+					// eslint-disable-next-line no-console
+					console.log( api.getTokenInfo( ids.tokenId ) );
+				} )
+				.on( 'mouseleave', function () {
+					// Deactivate all spans.
+					$( '.mw-parser-output .editor-token' ).removeClass( 'active' );
+				} );
+			components.widget.setState( 'ready' );
+		},
+		// Error handler.
+		function ( errorCode ) {
+			components.widget.setState( 'err' );
+			components.widget.setErrorMessage( errorCode );
+		}
+	);
 components.widget.on( 'close', () => {
 	// Close button; revert back to the original content
 	// $( '.mw-parser-output' ).empty().append( components.$originalOutput );
--- a/src/less/general.less
+++ b/src/less/general.less
@@ -1,0 +1,3 @@
+.mw-parser-output .editor-token.active {
+	background-color: #facc48;
+}


i18n/en.json
```diff
@@ -8,5 +8,7 @@
 	"ext-whowrotethat-state-error": "Error: $1",
 	"ext-whowrotethat-state-error-generic": "Could not load WhoWroteThat. Please refresh and try again.",
 	"ext-whowrotethat-ready-title": "Who Wrote That?",
-	"ext-whowrotethat-ready-general": "Hover to see contributions by the same author. Click for more details."
+	"ext-whowrotethat-ready-general": "Hover to see contributions by the same author. Click for more details.",
+
```
mooeypoo: Delete empty line

i18n/qqq.json
```diff
@@ -8,5 +8,7 @@
 	"ext-whowrotethat-state-error": "Displays in the information bar in case there were errors while loading.",
 	"ext-whowrotethat-state-error-generic": "A generic error suggesting to the user that they should refresh.",
 	"ext-whowrotethat-ready-title": "Title of the extension that appears when the system is loaded and displays information about the selection.",
-	"ext-whowrotethat-ready-general": "Instruction label that appears when the system is ready, before any hover or selection was made yet."
+	"ext-whowrotethat-ready-general": "Instruction label that appears when the system is ready, before any hover or selection was made yet.",
+
```
mooeypoo: Delete empty line

i18n/en.json
```diff
@@ -8,5 +8,7 @@
...
+	"ext-whowrotethat-requested-data-is-not-currently-available-in-wikiwho-database-it-will-be-available-soon": "WikiWho data is not currently available for this article. Please try again soon."
```
mooeypoo: That's....... a tad long for a key 😝 

Let's make it something more standardized, like `ext-whowrotethat-api-error-wikiwho-generic` ? or something like that?

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
 	 *
-	 * @param  {string} wikiUrl URL of the wiki page that we want to analyze.
-	 * @return {Promise}
+	 * @param {string} wikiUrl URL of the wiki page that we want to analyze.
 	 */
 	getData( wikiUrl ) {
-		return $.getJSON( this.getAjaxURL( wikiUrl ) );
+		$.getJSON( this.getAjaxURL( wikiUrl ) ).then( ( function ( result ) {
+			// Handle error.
+			if ( !result.success && result.info && this.errorHandler ) {
```
mooeypoo: Are all errors on API success? Are there any that should be invoked with `.fail()` ?

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
+				// Turn the English error message into a i18n message key.
+				// @TODO Remove this when the WikiWho API returns useful error codes.
+				let errorCode = result.info
+					.toLowerCase()
+					.replace( / /g, '-' )
+					.replace( /[^a-z-]/g, '' );
+				// Possible messages:
+				// eslint-disable-next-line
+				// ext-whowrotethat-requested-data-is-not-currently-available-in-wikiwho-database-it-will-be-available-soon
+				this.errorHandler( 'ext-whowrotethat-' + errorCode );
```
mooeypoo: Eh, I don't think we should do this. A couple of reasons:

- As I wrote above in the i18n, it results in extremely long keys
- In general, best practice is to avoid auto-generating message keys, since it's hard to fetch where they're used when you want to fix them. You're doing well by adding a comment with the key, but if we auto-generate we risk having a ton of generated keys eventually, which will make it unweildy.
- There are multiple cases where we might want to use the same message for several different errors. If we do it this way, we'll have to duplicate the translations for the translators and have redundant keys.
- If the text changes from the wikiwho API, our entire translation keys change, which is a mess for translators, worse than if only the message itself changes with the key remaining the same :\

Since we're already looking at the returning text, I think the better way to do this is to assign, ourselves, either error codes or error shorthands. You can still "normalize" the string we're getting so we don't mess with punctuation on those things, but simplify them to generic codes. For example, doing something like
```
if ( errorCode.toLower() === 'requested-data-is-not-currently-available-in-wikiwho-database-it-will-be-available-soon' ) {
 this.errorHandler( '100', 'wikiwho-generic' )
}
```
(See my comment below about rejecting the promise and returning the error code with the rejection)

Or something like that. Then as a followup (another .then perhaps) we go with a switch that can easily group some of those together for the same message, or output a message for whatever is relevant.

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
+			};
+
+			// Reorganise the revisions to group by user ID.
+			// revisions structure: {rev_id: [timestamp, parent_rev_id, user_id, editor_name ], ...}
+			for ( let [ revId, rev ] of Object.entries( result.revisions ) ) {
+				let userId = rev[ 2 ];
+				if ( cleanedResults.revisionsByUser[ userId ] === undefined ) {
+					cleanedResults.revisionsByUser[ userId ] = {};
+				}
+				cleanedResults.revisionsByUser[ userId ][ revId ] = {
```
mooeypoo: (I'm not sure about this, checking) -- when we have a token, do we have userID or revID on it? Do we need the data structure organized by the user with all their revs, or by rev with whatever data (userid,etc) for easy fetching?

I suspect you're organizing correctly, but I also suspect we might want a way to check directly on a revision (?) if that's true, then even adding another map that holds a reference to the same data could help. I might be wrong here, though, or it might not be needed just yet?

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
-	 * Get the WhoColor data for a given wiki page.
+	 * Get the WikiWho data for a given wiki page. Use onError and onResults to set return handlers.
 	 *
-	 * @param  {string} wikiUrl URL of the wiki page that we want to analyze.
-	 * @return {Promise}
+	 * @param {string} wikiUrl URL of the wiki page that we want to analyze.
 	 */
 	getData( wikiUrl ) {
-		return $.getJSON( this.getAjaxURL( wikiUrl ) );
+		$.getJSON( this.getAjaxURL( wikiUrl ) ).then( ( function ( result ) {
```
mooeypoo: if you return this promise, you won't need to do the trick of attaching `.onError`

So, if the process here cleans up the results, you could return the promise here, and then the process on `app.js` can be clean. 

Something like this in API:

```
return $.getJSON( ... )
    .then(
        // success
       function () {
             // On error, return a rejected promise with the error code/name/whatever we choose
             return $.Deferred().reject( errorCode );
             // On success, return the new cleaned up data
             return cleanedResults;
       },
       // API error
       function () {
            // Any API error like timing out, etc, should pop up in the failure state, so it can be processed to give
            // us some errorCode
            return errorCode
       }
     );
```

Then, in `app.js` you can "just" do this:

```
components.api.getData( ... )
    .then(
         function ( goodData ) {
               // Toggle the DOM, replace the HTML, etc etc
         },
         function ( errorCode ) {
              // Alll error codes, including the ones that started with supposed success
             // are now rejected promise, which makes it easier for us to now treat this as
             // a regular/common standardized error and display a message key in the InfoBarWidget
       }
);
```

This can make this much cleaner **and** separate concerns a lot more; it would mean that the API class deals with whatever funky ways wikiwho displays errors (even if the promise succeeded, we force-reject it, so the output from the API class gives us standard and cleaner way to handle success vs fails)

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
+					score: parseFloat( user[ 2 ] ).toFixed( 1 )
+				};
+			} );
+
+			this.resultHandler( cleanedResults );
+		} ).bind( this ) );
+	}
+
+	onError( callback ) {
+		this.errorHandler = callback;
```
mooeypoo: This has a bit of a code-smell of allowing (or needing) an external thing to deal with internal workings of how the API (and errors) are returned. See my comment above regarding returning the promise in `.getData`

Same for `onResult`

src/app.js
```diff
@@ -25,6 +25,36 @@ if ( !components ) {
...
+			// Revisions are: result.revisionsByUser[ userId ]
+			// and user info is: result.users[ userId ]
+		},
+		function () {
+			$( '.mw-parser-output .editor-token' ).removeClass( 'active' );
+		}
+	);
+	components.widget.setState( 'ready' );
+} );
+components.api.getData( window.location.href );
```
mooeypoo: To continue my point from the comment above, if we assume `getData` always returns a promise that is resolved or rejected based on errors, the flow of this code would also look nicer.

```
components.api.getData( ... )
   .then( ... ) // success
   .fail( ... ) // failure
```
Or it would also let us chain responses; for example, do something like 

```
showSpinner()
api.getData()
    .then( ... )
    .fail( ... )
    .always( () => hideSpinner );
```

mooeypoo (changes requested):
Overall this is taking shape! 

Small comments regarding some of the future-proofing of this, and some operational stuff with thought of separation of concerns.

Also, if you split out the analysis bit into its own method, you could run tests on it with some mock data...

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
+			result.present_editors.forEach( function ( user ) {
+				cleanedResults.users[ user[ 1 ] ] = {
+					username: user[ 0 ],
+					userId: user[ 1 ],
+					score: parseFloat( user[ 2 ] ).toFixed( 1 )
+				};
+			} );
+
+			this.resultHandler( cleanedResults );
+		} ).bind( this ) );
```
MusikAnimal: turn line 100 into a fat arrow `.then( result => {` then we don't need `.bind(this)` :)

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
+
+			// Reorganise the revisions to group by user ID.
+			// revisions structure: {rev_id: [timestamp, parent_rev_id, user_id, editor_name ], ...}
+			for ( let [ revId, rev ] of Object.entries( result.revisions ) ) {
+				let userId = rev[ 2 ];
+				if ( cleanedResults.revisionsByUser[ userId ] === undefined ) {
+					cleanedResults.revisionsByUser[ userId ] = {};
+				}
+				cleanedResults.revisionsByUser[ userId ][ revId ] = {
+					revId: revId,
```
MusikAnimal: not important but you could use just `revId,` (when the key/value share the same name)

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
+	 * @param {string} wikiUrl URL of the wiki page that we want to analyze.
 	 */
 	getData( wikiUrl ) {
-		return $.getJSON( this.getAjaxURL( wikiUrl ) );
+		$.getJSON( this.getAjaxURL( wikiUrl ) ).then( ( function ( result ) {
+			// Handle error.
+			if ( !result.success && result.info && this.errorHandler ) {
+				// Turn the English error message into a i18n message key.
+				// @TODO Remove this when the WikiWho API returns useful error codes.
+				let errorCode = result.info
```
MusikAnimal: this should probably a be a `const`

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
+					.replace( / /g, '-' )
+					.replace( /[^a-z-]/g, '' );
+				// Possible messages:
+				// eslint-disable-next-line
+				// ext-whowrotethat-requested-data-is-not-currently-available-in-wikiwho-database-it-will-be-available-soon
+				this.errorHandler( 'ext-whowrotethat-' + errorCode );
+				return;
+			}
+
+			let cleanedResults = {
```
MusikAnimal: This could also be a `const` to prevent re-declaration (insignificant here, but can help prevent bugs in larger applications)

src/Api.js
```diff
@@ -89,13 +92,68 @@ class Api {
...
+					revId: revId,
+					timestamp: rev[ 0 ],
+					parentRevId: rev[ 1 ],
+					userId: rev[ 2 ]
+				};
+			}
+
+			// Round the authorship percentages.
+			// present_editors structure: [ [ username, user_id, score ], ... ]
+			result.present_editors.forEach( function ( user ) {
```
MusikAnimal: `.forEach( user => {`

src/app.js
```diff
@@ -25,6 +25,36 @@ if ( !components ) {
 
 components.widget.toggle( true );
 components.widget.setState( 'pending' );
+components.api.onError( function ( errorCode ) {
```
MusikAnimal: Let's stick to ES6 if we can, `.onError( errorCode => {`

src/app.js
```diff
@@ -25,6 +25,36 @@ if ( !components ) {
...
+	components.widget.setState( 'err' );
+	components.widget.setErrorMessage( errorCode );
+} );
+components.api.onResult( function ( result ) {
+	// Insert modified HTML.
+	$( '.mw-parser-output' ).replaceWith( $( result.replacementHtml ) );
```
MusikAnimal: `$( '.mw-parser-output' ).html( result.replacementHtml )` might be cleaner

i18n/en.json
samwilson: Done.

i18n/qqq.json
samwilson: Done.

i18n/en.json
samwilson: Done.

src/Api.js
samwilson: I started by doing it that way, but wasn't sure how to pass the cleaned data through to the next success handler (i.e. `goodData`). I've rewritten to use a member variable of Api; is that correct?

src/Api.js
samwilson: Fixed with the below `.fail()` handling.

src/Api.js
samwilson: Removed anyway.

src/Api.js
samwilson: No, I was misunderstanding things; see the latest code for a better (maybe!) structure.

src/Api.js
samwilson: Code removed.

src/Api.js
samwilson: Good point. I think the code using Api can sort this out maybe. It feels like more of a UI matter.

src/app.js
samwilson: Done.

src/app.js
samwilson: Done.

samwilson (commented):
Thanks! All good points. I've rather rewritten it all now. See what you think. The error stuff is probably the biggest unresolved bit, but I'm not sure if we need to sort that out in this patch.

src/Api.js
```diff
@@ -89,13 +95,97 @@ class Api {
...
+	 * Get the WikiWho data for a given wiki page.
+	 *
+	 * @param {string} wikiUrl URL of the wiki page that we want to analyze.
 	 * @return {Promise}
 	 */
 	getData( wikiUrl ) {
-		return $.getJSON( this.getAjaxURL( wikiUrl ) );
+		const api = this,
+			request = $.getJSON( this.getAjaxURL( wikiUrl ) );
+		request.done( function ( result ) {
```
mooeypoo: The problem of using "done" is that you're not supposed to chain more 'thens' after it.

Whatever will take the returned request will not be able to consistently attach a .then; since this is meant for others to chain `then` and `fail` responses, you should change it to a `then` instead of done.

src/Api.js
```diff
@@ -89,13 +95,97 @@ class Api {
...
+			// Handle error response.
+			if ( !result.success && result.info && this.errorHandler ) {
+				// The API gives us an error message, but we don't use it because it's only
+				// in English.
+				return $.Deferred().reject( 'ext-whowrotethat-api-error-wikiwho-generic' );
+			}
+			// Store all results.
+			api.results = result;
+		} );
+		return request;
```
mooeypoo: You can just return directly, but it isn't crucial.

```
return $.getJSON( this.getAjaxURL( wikiUrl ) )
     .then( result => { ... } )
```

It's not crucial though.

src/Api.js
```diff
@@ -89,13 +95,97 @@ class Api {
...
+			request = $.getJSON( this.getAjaxURL( wikiUrl ) );
+		request.done( function ( result ) {
+			// Handle error response.
+			if ( !result.success && result.info && this.errorHandler ) {
+				// The API gives us an error message, but we don't use it because it's only
+				// in English.
+				return $.Deferred().reject( 'ext-whowrotethat-api-error-wikiwho-generic' );
+			}
+			// Store all results.
+			api.results = result;
```
mooeypoo: Why not return a construct here with the necessary items for the `.then` / `.fail`?

I guess it's a style choice and it's up to you -- we either treat `getData()` as a getter of state and always expect a promise (if it was previously rejected or resolved then attaching a .then to it runs immediately) **or** we treat it as a marker to say that the data has been fetched (which is what you're doing). Both are valid approaches, I marginally prefer the consistent promise but that's because I love promises and find the expectation consistent :stuck_out_tongue_closed_eyes: It's not crucial either way.

Whatever we do choose, we should make sure that the overlaying code has unified expectations (see my comment on the `getData` method call about caching or flagging)

src/Api.js
```diff
@@ -89,13 +95,97 @@ class Api {
...
+		};
+	}
+
+	/**
+	 * Get the WikiWho data for a given wiki page.
+	 *
+	 * @param {string} wikiUrl URL of the wiki page that we want to analyze.
 	 * @return {Promise}
 	 */
 	getData( wikiUrl ) {
```
mooeypoo, line 173: We need to do somethign here to allow for repeated calls to this function, since it's called from `app.js` which is triggered every time the `WhoWroteThat` trigger button is clicked.

My thought is that we could cache this promise, and return it with all the needed data. The system is supposed to allow for toggling, so it's conceivable that a user will click the "Who Wrote That?" button multiple times (initially it will load, then the user can click the [x] on the InfoBarWidget which resets back to the previous content, and then click the "Who Wrote That?" trigger again -- at this point, we should return the existing data from the API rather than re-request it)

Every time, it will call `getData()` -- if we cache the promise, it will mean that the first time the promise will run, and then resolve or reject -- and whatever in `app.js` will then run the `.then` or `.fail` on.

Subsequent clicks will run `getData()` but the promise will be cached -- which means it's already resolved (or rejected) which is a no-op, returning the original data, and the `then` and `fail` will act the same way.

Alternatively, we cache the state ("finished" / "ready" / etc) and use `getData()` just as a marker for finishing the operation (which is somewhat what you're doing now, since you're storing the results in class variables that can be called, so you just need to make sure they're ready.)

Either way, we have to make sure that `getData()` doesn't produce unnecessary API calls on subsequent clicks.

If we use a flag technique, then you can fake the return if the flag has already settled. For example,

```
getData( wikiUrl ) {
    if ( this.successful ) {
         return $.Deferred.resolve();
    } else if ( this.error ) {
         return $.Deferred.reject( this.error );
    }
    // The rest of the code, actually fetching
}
```

Otherwise, you should cache the promise and return it if it exists; something like
```
getData( wikiUrl ) {
   if ( this.promise ) {
         return this.promise;
   }
   this.promise = $.getJSON( ... ).then( .... );

   return this.promise;
}
```

mooeypoo (changes requested):
This is looking much nicer!

src/Api.js
mooeypoo: To clarify: Attaching a .then will work in the sense that the code inside that `then` or `fail` would *run*, but it has no promise of order like a chained `then` does. If we continue with using this promise as a sort of flag ("we got the data, doesn't matter if the rest of the stuff works later/earlier") it might matter less, but if we ever need to order the responses, it may matter more.

src/Api.js
```diff
@@ -89,13 +95,97 @@ class Api {
...
+	 * `class='token-editor-Y'` attributes.
 	 *
-	 * @param  {string} wikiUrl URL of the wiki page that we want to analyze.
+	 * @param {HTMLElement} element
+	 * @return {int[]|boolean[]} A two-element array: 0 = token_id, 1 = user_id.
+	 */
+	getIdsFromElement( element ) {
+		const out = [ false, false ],
+			tokenMatches = element.getAttribute( 'id' ).match( /token-(\d+)/ ),
+			editorMatches = element.getAttribute( 'class' ).match( /token-editor-(\d+)/ );
```
MusikAnimal: you could use `element.id` and `element.className` instead of using `getAttribute`

samwilson: Done.

samwilson: Done.

src/Api.js
samwilson: I've gone with promise-caching approach.

src/Api.js
samwilson: Makes sense. I've fixed it up.

src/Api.js
samwilson: The handler in app.js doesn't actually need the results, it just calls the other methods of Api that work with the stored results.

src/Api.js
samwilson: Switched to using a member var for the return value, for caching.

samwilson (commented):
Thanks both of you! I'm sorry this is such confused code. Hopefully it's getting closer. :-)

src/Api.js
mooeypoo: > The handler in app.js doesn't actually need the results, it just calls the other methods of Api that work with the stored results.

Yeah, that's how es5-style JS singleton was done (see the activation.singleton.js file I'm overriding with this patch) -- the methods are all private, and you have an exposed method that always gives you the same instance, then calls whatever method is needed.

One of the issues with a singleton is that it's hard to do tests on it, because it's not "pure"; it's always polluted by the previous tests. I ran into this in here too which is why the tests are in this order (if you move the first test to be the second, it'll fail :\ ) I almost made the ActivationSingleton into just a function that shares an instance, but in the *actual* operation of the system it *is* supposed to be a Singleton, so I think a small payment of ordering the tests might be worth it.

src/Api.js
mooeypoo: I think the UI should be the one picking what actual mw message to display, but the API class should simplify the errors into manageable codes/strings/whatever.

So the API doesn't have to deal with actual message keys, it can "just" output some base code -- the UI then has some `switch` statement based on what product tells us, that goes "error codes x, y get message key `blah`, error code z gets message key `foo`" and we can change whatever displays what in the UI -- but for that, the API needs to simplify them.

You're doing that in the next version anyways, but I think we might want to polish some of it once we implement the actual error message handling in the other ticket.

src/Api.js
mooeypoo: You can use a member variable, especially since it looks like the getData method is more about "wait until I'm done" rather than "let me give you the data you need".

In general, though, in order to deliver the results onwards, you "just" need to return them from within the promise. So, in this case, you could just `return goodData;` at the end. This would mean that whatever attaches next to this promise, has that parameter for the success --

```
api.getData( ... ).then( result => { /* result is the "goodData" you returned */ } );
```

This is more for general future/knowledge/discussion. I think the strategy you go with is also fine, and means there's no need to return the values. I wrote above that I have a preference towards doing everything in the promise (returning the values and using them as your starting point) but that, too, has some edge case price, so it's not really crucial which way we choose. I think the way you went with is good.

src/Api.js
```diff
@@ -89,13 +91,100 @@ class Api {
...
+		const out = [ false, false ],
+			tokenMatches = element.id.match( /token-(\d+)/ ),
+			editorMatches = element.className.match( /token-editor-([^\s]+)/ );
+		if ( tokenMatches && tokenMatches[ 1 ] ) {
+			out[ 0 ] = parseInt( tokenMatches[ 1 ] );
+		}
+		if ( editorMatches && editorMatches[ 1 ] ) {
+			out[ 1 ] = editorMatches[ 1 ];
+		}
+		return out;
```
mooeypoo, line 110: Is there a reason why we want to keep using arrays rather than a readable object with keys?

It's not really crucial, but since this is what **we** make (rather than what we get from the WikiWho api) we can start making things more readable for maintainability and readability.

I don't think this is a merge-preventer, but I noticed it and thought we should consider these things. The fact that WikiWho gives us unintelligible array-stuff made me a bit more sensistive to arrayed results :p

src/Api.js
```diff
@@ -89,13 +91,100 @@ class Api {
...
+	 */
+	getReplacementHtml() {
+		return this.results.extended_html;
+	}
+
+	/**
+	 * Get user and revision information for a given token.
+	 *
+	 * @param {int} tokenId
+	 * @return {{revisionId: *, score: *, userId: *, username: *, revisionTime: *}|boolean}
```
mooeypoo: Write a bit more explanation.

Returns an object that represents the token info (breakdown) or false if a token wasn't found.

src/Api.js
```diff
@@ -89,13 +91,100 @@ class Api {
...
+	 */
+	getTokenInfo( tokenId ) {
+		let revId, revision, username, score;
+
+		// Get the token information. results.tokens structure:
+		// [ [ conflict_score, str, o_rev_id, in, out, editor/class_name, age ], ... ]
+		// e.g. Array(7) [ 0, "indicate", 769691068, [], [], "18201938", 76652371.587203 ]
+		const token = this.results.tokens[ tokenId ] || false;
```
mooeypoo: Why do you need "false" here? You're later checking `if ( !token )` which would trigger for false, empty string, 0 or `undefined`s anyways.

src/Api.js
```diff
@@ -89,13 +91,100 @@ class Api {
...
+			revisionTime: new Date( revision[ 0 ] ),
+			score: score
+		};
+	}
+
+	/**
+	 * Get the WikiWho data for a given wiki page.
+	 *
+	 * @param {string} wikiUrl URL of the wiki page that we want to analyze.
 	 * @return {Promise}
```
mooeypoo: Add info about return; something like `returns a promise that resolves when the data is ready, or rejects if there was an error`

This is a bit more important since we're not returning data with the promise, it's meant to serve as a sort of flag/wait until the data's available for us to call the other class methods.

src/app.js
```diff
@@ -25,6 +25,37 @@ if ( !components ) {
...
+					// Deactivate all spans.
+					$( '.mw-parser-output .editor-token' ).removeClass( 'active' );
+				}
+			);
+			components.widget.setState( 'ready' );
+		},
+		// Error handler.
+		function ( errorCode ) {
+			components.widget.setState( 'err' );
+			components.widget.setErrorMessage( errorCode );
```
mooeypoo, line 55: We can clean that up later, but this is where we can take the error codes and connect them to mw.msg keys so that the API doesn't need to know or output actual keys.

We have another ticket about sorting out the actual error messaging, this refinement can go there, but since we discussed it, I wanted to point out to what I meant.

mooeypoo (changes requested):
This is almost ready; I delved a little deeper with some relatively minor change requests. I think we're almost ready to merge this thing and move on to the UI work :clap:

src/Api.js
samwilson: Good point; I've no idea what I was thinking. :) Fixed.

src/Api.js
samwilson: Removed.

src/Api.js
samwilson: Done.

src/app.js
samwilson: Yeah, I'm figuring that this whole bit of code will be refactored; it's mostly a proof of concept so far.

src/app.js
```diff
@@ -25,6 +25,37 @@ if ( !components ) {
...
+components.api.getData( window.location.href )
+	.then(
+		// Success handler.
+		function () {
+			const api = components.api;
+			// Insert modified HTML.
+			$( '.mw-parser-output' ).html( api.getReplacementHtml() );
+			// Highlight when hover a user's contributions.
+			// @TODO This is just testing code and should be replaced by the proper behaviour.
+			$( '.mw-parser-output .editor-token' ).hover(
```
MusikAnimal: `JQMIGRATE: jQuery.fn.hover() is deprecated`

MusikAnimal (approved):
LGTM!

mooeypoo (approved):

PR #27, opened 2019-08-15
Simplify activation routine, and unify as ES6

mooeypoo:
- Convert the activation singleton to an ES6 class
- Load dependencies and store original content for reuse
- Add tests for the activation singleton
- Make sure the app loads consistently, within the dependency load


src/ActivationSingleton.js
```diff
@@ -0,0 +1,160 @@
...
+		// whether messages are set properly, since that
+		// has its own tests in the mw.messsages bundle
+		if ( window.mw ) {
+			// Load all messages
+			mw.messages.set(
+				$.extend( {},
+					// Manually create fallback on English
+					this.translations.en,
+					this.translations[ this.lang ]
+				)
```
MusikAnimal, line 97: I would recommend using `Object.assign` over `$.extend`

src/App.js
```diff
@@ -0,0 +1,87 @@
...
+	 *
+	 * @constructor
+	 * @param {App} A class instance
+	 */
+	constructor() {
+		// Instantiate only once
+		if ( !App.instance ) {
+			this.initialized = false;
+			App.instance = this;
+		}
```
MusikAnimal, line 25: This system (here and in ActivationSingleton.js) feels a little weird. Instead of preventing reinstantiation within this class, we could prevent it wherever the class is being used, no?

I'm sure I'm missing something, and frankly it seems kind of neat to not have to worry about re-using an existing instance. It just deviates from what I would normally expect a constructor to do (`new Whatever()` should always give me a new `Whatever`)

MusikAnimal (commented):
Looks good to me, I think! I left some feedback, but I'm not necessarily requesting changes.

src/App.js
mooeypoo: Yeah, this is why it's a singleton; a Singleton only allows for one instantiation, no matter what. 
https://en.wikipedia.org/wiki/Singleton_pattern

So the idea here is that you can import the file and know that you're always using the same instance.

Thta said, I think I can make it by having a shared instance, rather than forcing a singleton. It's more "proper" as a singleton, but I don't know if it's completely needed here. I'll take another look to see if the other method is nicer while keeping up with the needed spec.

mooeypoo: BTW, this is a super awesome writeup about Javascript design patterns: https://addyosmani.com/resources/essentialjsdesignpatterns/book/

it was written for ES5, but the conceptual design patterns still apply. The code that I used is just the way to do a singleton for an ES6 class.

samwilson: In other languages, a singleton would have a private constructor, and some factory method. I guess JS can't do that, so ensuring the constructor only ever returns the same instance is as good as we get.

mooeypoo: Yeah, that's how es5-style JS singleton was done (see the activation.singleton.js file I'm overriding with this patch) -- the methods are all private, and you have an exposed method that always gives you the same instance, then calls whatever method is needed.

One of the issues with a singleton is that it's hard to do tests on it, because it's not "pure"; it's always polluted by the previous tests. I ran into this in here too which is why the tests are in this order (if you move the first test to be the second, it'll fail :\ ) I almost made the ActivationSingleton into just a function that shares an instance, but in the actual operation of the system it is supposed to be a Singleton, so I think a small payment of ordering the tests might be worth it.

src/ActivationSingleton.js
mooeypoo: Ooh, good point. Done.

Gruntfile.js
```diff
@@ -145,60 +145,8 @@ module.exports = function Gruntfile( grunt ) {
...
-				// Done async
-				done();
-			}
-		);
-	} );
-
 	grunt.registerTask( 'lint', [ 'eslint', 'stylelint', 'banana' ] );
 	grunt.registerTask( 'test', [ 'lint', 'run:tests' ] );
-	grunt.registerTask( 'build', [ 'clean', 'less', 'generateProductionScript' ] );
+	grunt.registerTask( 'build', [ 'clean', 'less', 'replace:language', 'browserify' ] );
```
samwilson: The `copy:browserextension` needs to be added here.

src/ActivationSingleton.js
```diff
@@ -0,0 +1,161 @@
...
+			);
+
+			// Add a portlet link to 'tools'
+			this.link = mw.util.addPortletLink(
+				'p-tb',
+				'#',
```
samwilson, line 103: This makes it jump to the top now (it didn't do this before).

samwilson (changes requested):

Gruntfile.js
mooeypoo: Added `copy` in general in there.

src/ActivationSingleton.js
mooeypoo: Oops, I took out `e.preventDefault();` and forgot to put it back in. Fixed in the new one.

samwilson:
Why's there no 'delete branch' button here now?

PR #28, opened 2019-08-16
Remove generated files and reorganize others

samwilson:
* Make the dist/ and temp/ directories only for generated files;
* Remove generated files from Git;
* Add JQuery from node_modules, and correct its version to what we were using; and
* Move grunt clean task to beginning, to ensure builds start in a known state.


Gruntfile.js
```diff
@@ -123,7 +124,16 @@ module.exports = function Gruntfile( grunt ) {
 				]
 			}
 		},
-		clean: [ generatedFile ],
+		copy: {
+			browserextension: {
+				files: [
+					{ src: 'build/extension_manifest.json', dest: 'dist/extension/manifest.json' },
```
mooeypoo, line 130: ooh, smart!

.gitignore
```diff
@@ -1,7 +1,6 @@
 .DS_STORE
 node_modules/
-temp/*
-!temp/.gitkeep
```
mooeypoo, line 4: Do we not want those two folders to appear in the structure (without files in them) as placeholders?

mooeypoo (approved):
Looks good to me... and we'll soon make the gruntfile even simpler, so the build process will be (hopefully) cleaner in general!

PR #30, opened 2019-08-16
Remove jQuery dependency from extension initialization

mooeypoo:
We actually don't need jQuery in the extension dependencies.
We are injecting the full script into the MediaWiki DOM, and use
the jquery that exists within that ecosystem. Our 'inject' mechanism
doesn't require jQuery itself, so there's no need to load jQuery
in the manifest.json.


MusikAnimal (approved):
Still works in my testing

PR #32, opened 2019-08-20
Add welcome popup for first-time use

mooeypoo:
- Add a welcome popup when the extension is active for the first time
- Only show if the popup was not previously dismissed
  - For the extension, the popup script is only injected at all if it is needed and used.
  - For the gadget, the script is included but the popup is only triggered if it was not previously dismissed.
- Store dismissed state:
  - For the extension, use browser global storage
  - For the gadget, use localStorage
- Add a debugging namespace `wwtDebug`, with a method that allows for resetting the state of the popup dismissal for testing.
- Update the README.md file to show the new `wwtDebug` usage

### DIFF
--- a/build/extension_content_script.js
+++ b/build/extension_content_script.js
@@ -1,3 +1,4 @@
+var theBrowser = chrome || browser;
 /**
  * injectScript - Inject internal script to available access to the `window`
  *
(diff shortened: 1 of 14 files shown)


build/extension_manifest.json
```diff
@@ -30,6 +30,6 @@
 		"css": [ "generated.whowrotethat.css" ],
 		"run_at": "document_end"
 	} ],
-	"web_accessible_resources": [ "js/generated.pageScript.js" ],
-	"permissions": [ "activeTab" ]
+	"web_accessible_resources": [ "js/generated.pageScript.js", "js/generated.welcomeTour.js" ],
+	"permissions": [ "activeTab", "storage" ]
```
samwilson: To get this working as a temporary addon in Firefox I had to add the following to `manifest.json`:

	"browser_specific_settings": {
		"gecko": {
			"id": "whowrotethat@wikimedia"
		}
	},

The value of `id` doesn't matter, it just had to be set.

build/extension_content_script.js
```diff
@@ -14,8 +15,62 @@ function injectScript( filePath, tag ) {
...
 }
 
+/**
+ * Consider whether to activate the initial welcome tour
+ * and listen to dismissal event if so. If dismissal event
+ * happened, store that state so the welcome tour is never
+ * shown again.
+ */
+function activateWelcomeTour() {
+	theBrowser.storage.sync.get( [ 'WelcomeTourSeen' ], function ( result ) {
```
samwilson, line 25: I think there might be some documentation confusion about this, because [according to MDN](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/get) this should be more like `.get( [ 'WelcomeTourSeen' ] ).then( function ( result ) { ... } );` — but the callback as the 2nd arg seems to work too!

I'd still say maybe go with the Promise structure, because then it's also possible to add an error handler in (which is what I did to find the addon ID issue I was having).

i18n/qqq.json
```diff
@@ -11,5 +11,9 @@
 	"ext-whowrotethat-error-contact": "Error message shown if an error is unlikely to be recoverable, with a link to the Who Wrote That contact page.\n* $1 is the HTML link to the contact page, with the text specified in {{wm-msg|ext-whowrotethat-state-error-contact-link}}",
 	"ext-whowrotethat-error-contact-link": "The text of the link to the contact page shown in {{wm-msg|ext-whowrotethat-state-error-contact}}.",
 	"ext-whowrotethat-ready-title": "Title of the extension that appears when the system is loaded and displays information about the selection.",
-	"ext-whowrotethat-ready-general": "Instruction label that appears when the system is ready, before any hover or selection was made yet."
+	"ext-whowrotethat-ready-general": "Instruction label that appears when the system is ready, before any hover or selection was made yet.",
+	"ext-whowrotethat-api-error-wikiwho-generic": "Error message displayed when an article's data could not be fetched.",
```
samwilson: Where's this error message used? I thought we'd got rid of it.

samwilson (changes requested):
I'm not sure if the errors I'm seeing are just related to my dev environment.

In addition to my comments below, I'm also seeing intermittent (!) layout issues with the popup, which really don't seem likely to be coming from the code here because it's just using standard OOUI... but I can't figure it out. You might be able to recreate by going to an article section in a new tab and then scrolling back up to the popup without clicking anywhere.

Arrow disconnected:

wwt offset

Popup is off screen to the left:

wwt offscreen

i18n/qqq.json
mooeypoo: Whoops, it's a rebase fluke

build/extension_manifest.json
mooeypoo: Done

build/extension_content_script.js
mooeypoo: Yeah promises are better. I think I was using what I saw in the Chrome docs, which was the callback... but promises are better.

mooeypoo:
Regarding the wrong placement -- this was weeeeird. I added a `$overlay` (good idea regardless to make sure positioning and zindex is right) and fixed it by setting `hideWhenOutOfView: false,`

This is a weird OOUI bug. The positioning should still work with the default `hideWhenOutOfView: true` but it seems to not do that very well when the element is out of view. bah.

build/extension_content_script.js
mooeypoo: As it turns out, Chrome doesn't do the promises. I reverted back to using a callback :\

samwilson (approved):

PR #33, opened 2019-08-20
Extension: Exclude special pages and log to console

mooeypoo:
Make sure the extension itself is not loading in any special page.
The code is already there to make the button not appear, but it is
best if the browser knows not to load the extension files at all
for excluded namespaces. The biggest and most problematic namespace
is the `Special:` namespace, that includes within it pages like
`Special:Login` and `Special:Preferences` where we don't want the
code to load at all.

In the case of a gadget, the MediaWiki system knows not to load
the gadget on "sensitive" pages (like `Special:Preferences` and
`Special:Login`) so those, at least, are covered, and the code
itself verifies that the button is not injected into the DOM in
pages that are not articles.

### DIFF
--- a/build/extension_manifest.json
+++ b/build/extension_manifest.json
@@ -13,6 +13,18 @@
 			"*://tr.wikipedia.org/*",
 			"*://es.wikipedia.org/*"
 		],
+		"exclude_globs": [
+			"*://en.wikipedia.org/wiki/Special:*",
+			"*://de.wikipedia.org/wiki/Spezial:*",
+			"*://eu.wikipedia.org/wiki/Berezi:*",
+			"*://tr.wikipedia.org/wiki/Özel:*",
+			"*://es.wikipedia.org/wiki/Especial:*",
+			"*://en.wikipedia.org/*title=Special:*",
+			"*://de.wikipedia.org/*title=Spezial:*",
+			"*://eu.wikipedia.org/*title=Berezi:*",
+			"*://tr.wikipedia.org/*title=Özel:*",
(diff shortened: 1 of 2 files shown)


MusikAnimal:
I still see `WhoWroteThat Extension: Loaded on page` on Special:UserLogin, etc. Is that expected?

Maybe Dom should test this PR since us both are unable to reproduce the bug he found.

mooeypoo:
> I still see `WhoWroteThat Extension: Loaded on page` on Special:UserLogin, etc. Is that expected?
> 
> Maybe Dom should test this PR since us both are unable to reproduce the bug he found.

O.O ... I don't........ I'm going to retest this again. Anything with Special:* should not load this extension at all, including the `console.log` output.

A couple of potential debugging help questions:

- Can you verify you've reloaded the extension in `chrome://extensions` after running `grunt build` with this patch?
- Can you share the complete URL of where you see this? This patch will block the extension on `wiki/Special:UserLogin` but might not on `/w/index.php?title=Special:UserLogin` ... is that what you see?

This is baffling.

build/extension_manifest.json
```diff
@@ -13,6 +13,13 @@
 			"*://tr.wikipedia.org/*",
 			"*://es.wikipedia.org/*"
 		],
+		"exclude_globs": [
+			"*://en.wikipedia.org/wiki/Special:*",
```
samwilson, line 17: Each of these probably also need `"*://en.wikipedia.org/*title=Special:*"` — don't think there's anything caught by that that shouldn't be.

mooeypoo: Added, with the translations.

PR #34, opened 2019-08-21
Improve error messages

samwilson:
Introduce three error states and show each of them according to
the state of the response from WikiWho.

https://phabricator.wikimedia.org/T226760

Bug: T226760

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -4,10 +4,12 @@
 	},
 	"ext-whowrotethat-activation-link": "Who Wrote That?",
 	"ext-whowrotethat-activation-link-tooltip": "Activate Who Wrote That?",
-	"ext-whowrotethat-state-pending": "<strong><em>Who Wrote That?</em></strong> is loading. This might take a while",
-	"ext-whowrotethat-state-error": "Error: $1",
-	"ext-whowrotethat-state-error-generic": "Could not load WhoWroteThat. Please refresh and try again.",
+	"ext-whowrotethat-state-pending": "<strong><em>Who Wrote That?</em></strong> is loading. This might take a while.",
+	"ext-whowrotethat-state-error": "API Error: $1",
+	"ext-whowrotethat-error-refresh": "Please refresh the page or try again later.",
+	"ext-whowrotethat-error-later": "Please try again later.",
+	"ext-whowrotethat-error-contact": "Please $1 about this issue.",
+	"ext-whowrotethat-error-contact-link": "contact us",
 	"ext-whowrotethat-ready-title": "Who Wrote That?",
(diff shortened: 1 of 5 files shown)


src/Api.js
```diff
@@ -178,13 +178,31 @@ class Api {
...
 					// The API gives us an error message, but we don't use it because it's only
-					// in English.
-					return $.Deferred().reject( 'ext-whowrotethat-api-error-wikiwho-generic' );
+					// in English. Some of the error messages are:
+					// * result.info: "Requested data is not currently available in WikiWho
+					//   database. It will be available soon."
+					// * result.error: "The article (x) you are trying to request does not exist in
+					//   english Wikipedia."
+					// We do add the full error details to the console, for easier debugging.
+					let errCode = 'contact';
```
mooeypoo: Pedantic comment: You could do

```
const errCode = ( result.info && result.info.match( /data is not currently available/i ) ?
     'refresh' : 'contact';
```

i18n/en.json
```diff
@@ -4,10 +4,12 @@
...
 	"ext-whowrotethat-activation-link-tooltip": "Activate Who Wrote That?",
-	"ext-whowrotethat-state-pending": "<strong><em>Who Wrote That?</em></strong> is loading. This might take a while",
-	"ext-whowrotethat-state-error": "Error: $1",
-	"ext-whowrotethat-state-error-generic": "Could not load WhoWroteThat. Please refresh and try again.",
+	"ext-whowrotethat-state-pending": "<strong><em>Who Wrote That?</em></strong> is loading. This might take a while.",
+	"ext-whowrotethat-state-error": "API Error: $1",
+	"ext-whowrotethat-state-error-refresh": "Please refresh the page or try again later.",
+	"ext-whowrotethat-state-error-later": "Please try again later.",
+	"ext-whowrotethat-state-error-contact": "Please $1 about this issue.",
+	"ext-whowrotethat-state-error-contact-link": "contact us",
```
mooeypoo: Super pedantic comment that you should feel free to ignore :stuck_out_tongue: --

I'd rename those messages to be something like `ext-whowrotethat-error-refresh` and `ext-whowrotethat-error-contact` etc because `ext-whowrotethat-state-*` are InfoWidget states in general, and the error messages you're adding is literal error messages.

... or I'm just wasting your time with this super pedantic silliness. It can go either way :laughing:

src/Api.js
samwilson: I was thinking that we will want to extend this with matching different error messages, but you're right, it could be simpler for the time being.

Done.

i18n/en.json
samwilson: I love pedantry!

Done.

mooeypoo (approved):

PR #39, opened 2019-08-29
Only load welcome popup on valid pages

mooeypoo:
While the extension only actually loads on valid pages, there are
cases (like talk page and history, etc) where the files themselves
are loaded but the script won't when the activation singleton
checks for valid pages.

However, that check is only done for initialization of the main
script, and not for the welcome popup.

More importantly, the gadget always loads the welcome popup, and
we have to make sure it does not run or display it for invalid
pages since the validity check was only done for the activation
initialization of the general script.

This commit adds that check to the welcome popup as well, so it
does not load if the page is not valid.

### DIFF
--- a/src/outputs/browserextension_tour.js
+++ b/src/outputs/browserextension_tour.js
@@ -7,6 +7,9 @@
 				'mediawiki.jqueryMsg'
 			] )
 		).then( function () {
+			// The check about whether the tour was already seen has already
+			// taken place at this point. This whole file would not be injected
+			// by the contentScript to begin with, if WelcomeTourSeen is true
 			if ( !$( '#t-whowrotethat' ).length ) {
 				return;
 			}
--- a/src/outputs/gadget.js
+++ b/src/outputs/gadget.js
@@ -33,8 +33,9 @@ import languageBlob from '../../temp/languages'; // This is generated during the
 
 		loadWhoWroteThat();
 
-		if ( welcomeTourSeen ) {
+		if ( welcomeTourSeen || !activationInstance.isValidPage() ) {
 			// Do not show the tour if it was previously dismissed
+			// Or if the page is invalid
 			return;
 		}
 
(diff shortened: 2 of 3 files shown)


src/outputs/browserextension_tour.js
```diff
@@ -7,7 +8,7 @@
 				'mediawiki.jqueryMsg'
 			] )
 		).then( function () {
-			if ( !$( '#t-whowrotethat' ).length ) {
+			if ( !$( '#t-whowrotethat' ).length || !activationInstance.isValidPage() ) {
```
samwilson: I'm a bit confused. Aren't we already using `isValidPage` to determine if we add `#t-whowrotethat`? So it already shouldn't be present on invalid pages.

mooeypoo: The extension, yes. The gadget - no.

I just thought that if I'm already adding that check to the gadget, I might as well make it consistent. It is, as you say, technically redundant here. Should I take it off? :\

src/outputs/gadget.js
```diff
@@ -33,8 +33,9 @@ import languageBlob from '../../temp/languages'; // This is generated during the
 
 		loadWhoWroteThat();
 
-		if ( welcomeTourSeen ) {
+		if ( welcomeTourSeen || !activationInstance.isValidPage() ) {
```
samwilson, line 36: I'm sure you're right here, but I can't quite figure it out myself. I would have said that this should be

    if ( welcomeTourSeen || !$( '#t-whowrotethat' ).length ) {

to be equivalent to what's done in `src/outputs/browserextension_tour.js`. (And that maybe the check in that file could happen in a similar place as this one, i.e. *before* loading the required modules... but that's probably a separate matter.)

mooeypoo: That's a good point, although it's also basically the exact same, since `$( '#t-whowrotethat' )` would only be injected to the DOM if `isValidPage` is true, so it's more a logical/flow thing.

I can switch it up for logic's sake.

PR #40, opened 2019-08-29
WelcomePopup: Fix positioning problem for long pages

mooeypoo:
When we have long pages where the "tools" menu is under the viewport,
the `<html>` tag, unhelpfully, stops (its height is limited) at the
edge of the viewport. This means that the `<body>` is similarly limited
in height, even though the content itself keeps going.

As a result, the popup, when instantiated, cannot be placed below
that point, which creates a popup that appears above where it should.

This is a hacky way to fix this. When and if the popup is active,
the code adds a class to the `html` block and sets it at a large
min-height that will cover at least the side menu to be able to
get the popup to appear properly.

HTML sucks.

### DIFF
--- a/.jsdoc.json
+++ b/.jsdoc.json
@@ -34,7 +34,8 @@
         "jQuery": "https://api.jquery.com/Types/#jQuery",
         "jQuery.Promise": "https://api.jquery.com/Types/#Promise",
         "Object": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object",
-        "Promise": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise"
+        "Promise": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise",
+        "OO.ui.PopupWidget": "https://doc.wikimedia.org/oojs-ui/master/js/#!/api/OO.ui.PopupWidget"
       }
     }
   }
--- a/src/WelcomePopupWidget.js
+++ b/src/WelcomePopupWidget.js
@@ -2,6 +2,7 @@
  * A popup meant to display as a welcome message after the
  * extension is installed for the first time.
  * @class
+ * @extends OO.ui.PopupWidget
  *
  * @constructor
  * @param {Object} [config={}] Configuration options
@@ -49,6 +50,24 @@ const WelcomePopupWidget = function WelcomePopupWidget( config = {} ) {
 
 /* Setup */
 OO.inheritClass( WelcomePopupWidget, OO.ui.PopupWidget );
+
+/* Methods */
+
+/**
+ * @inheritdoc
+ */
+WelcomePopupWidget.prototype.toggle = function ( show ) {
+	// Parent method
+	WelcomePopupWidget.parent.prototype.toggle.call( this, show );
+
+	// HACK: Prevent clipping; we don't want (or need) the
+	// popup to clip itself when it's outside the viewport
+	// or close to the viewport.
+	// Unfortunately, the toggle operation calls toggleClipping
+	// multiple times on and off, so we have to insist on it
+	// being actually off at the end of it.
+	this.toggleClipping( false );
+};
 
 module.exports = WelcomePopupWidget;
 export default WelcomePopupWidget;
--- a/src/less/general.less
+++ b/src/less/general.less
@@ -1,3 +1,13 @@
+html.ext-wwt-popup,
+html.ext-wwt-popup body {
+	// HACK: By default, the html and <body> only span
+	// as far down as the viewport, even though the content
+	// is longer. This makes the welcome popup appear in
+	// weird places if the "Tools" menu is beyond the bottom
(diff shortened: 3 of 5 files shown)


MusikAnimal:
I'm getting `Error handling response: TypeError: Cannot read property 'WelcomeTourSeen' of undefined` on master, so I can't test this.

That aside, I wonder if it makes more sense for Who Wrote That link to be under `#p-cactions` (the "More" menu). If the viewport is that tall such that you can't see the Tools menu on page load, would they even see the welcome popup?

mooeypoo:
> I'm getting `Error handling response: TypeError: Cannot read property 'WelcomeTourSeen' of undefined` on master, so I can't test this.

eeeek. Is this in Firefox or Chrome or both? I think I know what's happening but I might need to see how to duplicate and we need to fix this either way.

> That aside, I wonder if it makes more sense for Who Wrote That link to be under `#p-cactions` (the "More" menu). If the viewport is that tall such that you can't see the Tools menu on page load, would they even see the welcome popup?

Interesting question. We should ask Ilana.

MusikAnimal (commented):
I can reproduce the issue now and can confirm `min-height: 100em` fixes it. Why 100em, though? Is there a scenario where this would be too much or too little?

> > I'm getting `Error handling response: TypeError: Cannot read property 'WelcomeTourSeen' of undefined` on master, so I can't test this.
> 
> eeeek. Is this in Firefox or Chrome or both? I think I know what's happening but I might need to see how to duplicate and we need to fix this either way.

Chromium. It was quite bad because the WWT wouldn't load at all. I'll try to figure out how to reproduce it and create a ticket.

> > That aside, I wonder if it makes more sense for Who Wrote That link to be under `#p-cactions` (the "More" menu). If the viewport is that tall such that you can't see the Tools menu on page load, would they even see the welcome popup?
> 
> Interesting question. We should ask Ilana.

Yeah, now that I got the WelcomePopup working (and with this fix), I can confirm that I don't see it at all unless I scroll down. I think everyone would eventually find it, and it is sort of a "Tool" (hence belongs in that list of links), but p-cactions is a common placement for things like this too.

mooeypoo:
> I can reproduce the issue now and can confirm `min-height: 100em` fixes it. Why 100em, though? Is there a scenario where this would be too much or too little?

There probably is, but from my tests, it seemed like a safe enough space without being obtrusive if the page is too short.

> > > I'm getting `Error handling response: TypeError: Cannot read property 'WelcomeTourSeen' of undefined` on master, so I can't test this.
> > 
> > 
> > eeeek. Is this in Firefox or Chrome or both? I think I know what's happening but I might need to see how to duplicate and we need to fix this either way.
> 
> Chromium. It was quite bad because the WWT wouldn't load at all. I'll try to figure out how to reproduce it and create a ticket.

Yeah let's deal with this in a separate ticket/PR.

> > > That aside, I wonder if it makes more sense for Who Wrote That link to be under `#p-cactions` (the "More" menu). If the viewport is that tall such that you can't see the Tools menu on page load, would they even see the welcome popup?
> > 
> > 
> > Interesting question. We should ask Ilana.
> 
> Yeah, now that I got the WelcomePopup working (and with this fix), I can confirm that I don't see it at all unless I scroll down. I think everyone would eventually find it, and it is sort of a "Tool" (hence belongs in that list of links), but p-cactions is a common placement for things like this too.

I raised this to Ilana and she's created a ticket. For the moment, we can continue as if it's still going to stay in the same place, and if we change it later we can get rid of this hack.

samwilson (commented):
This puts the popup in the right place, but makes it shrink and get a scroll bar!

scrollypopup

It's definitely an improvement though.

mooeypoo:
Yes, that's a "feature" of ClippableElement.

I had a whole comment about how I can't turn Clippable off, and then I realized I can, although it's a pretty yucky hack. But it does work, and honestly, the fact it's a hack is not our fault -- this should be doable upstream.

See latest change.

.jsdoc.json
```diff
@@ -34,7 +34,8 @@
         "jQuery": "https://api.jquery.com/Types/#jQuery",
         "jQuery.Promise": "https://api.jquery.com/Types/#Promise",
         "Object": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object",
-        "Promise": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise"
+        "Promise": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise",
+		"OO.ui.PopupWidget": "https://doc.wikimedia.org/oojs-ui/master/js/#!/api/OO.ui.PopupWidget"
```
MusikAnimal: extraneous tab?

MusikAnimal (changes requested):
Working great now! And I love the `height: unset` trick. I wouldn't even call that a "hack".

Just one minor comment about spacing, otherwise looks good to me.

.jsdoc.json
mooeypoo: Oops

MusikAnimal (approved):

PR #41, opened 2019-08-29
Add revision details popup

MusikAnimal:
Bug: https://phabricator.wikimedia.org/T228803

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -7,17 +7,30 @@ class Api {
 	/**
 	* @param {Object} config
 	* @cfg config.url The WikiWho base URL.
+	* @cfg config.mwApi The mw.Api instance.
 	* @constructor
 	*/
 	constructor( config = {} ) {
-		this.url = config.url || '';
+		// Remove trailing slash from URL.
+		this.url = ( config.url || '' ).replace( /\/$/, '' );
 
-		// Remove trailing slash.
-		if ( this.url && this.url.slice( -1 ) === '/' ) {
-			this.url = this.url.slice( 0, -1 );
-		}
-
+		this.mwApi = config.mwApi;
 		this.results = null;
+	}
+
+	/**
+	 * Fetch core messages needed for the revision popup, etc., making them available to mw.msg().
+	 * This is called just after the script is first loaded, and the request completes very quickly,
+	 * so we shouldn't need to bother with promises and such.
+	 */
+	fetchMessages() {
+		this.mwApi.loadMessages( [
+			'parentheses-start',
+			'talkpagelinktext',
+			'pipe-separator',
+			'contribslink',
+			'parentheses-end'
+		] );
 	}
 
 	/**
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,7 @@
 import config from './config';
 import Api from './Api';
 import InfoBarWidget from './InfoBarWidget';
+import RevisionPopupWidget from './RevisionPopupWidget';
 import activationInstance from './ActivationSingleton';
 
 /**
@@ -24,6 +25,8 @@ class App {
 			App.instance = this;
 		}
 
+		this.revisionPopup = new RevisionPopupWidget();
+
 		return App.instance;
 	}
 
@@ -37,7 +40,13 @@ class App {
 		}
 
 		this.widget = new InfoBarWidget( { state: 'pending' } );
-		this.api = new Api( { url: config.wikiWhoUrl } );
+		this.api = new Api( {
+			url: config.wikiWhoUrl,
+			mwApi: new mw.Api()
+		} );
+
+		// Pull in necessary core messages.
+		this.api.fetchMessages();
 
 		this.widget.setState( 'pending' );
 		// Attach widget
@@ -49,6 +58,7 @@ class App {
 
 		// Attach events
 		this.widget.on( 'close', this.onWidgetClose.bind( this ) );
+
 		this.initialized = true;
 	}
 
@@ -60,8 +70,6 @@ class App {
 	 * on and off and on again.
 	 */
 	start() {
-		const self = this;
-
 		this.initialize();
 		this.widget.toggle( true );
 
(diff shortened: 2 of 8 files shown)


src/Api.js
```diff
@@ -12,6 +12,11 @@ class Api {
 	constructor( config = {} ) {
 		this.url = config.url || '';
 
+		// This validation is for tests, where mw is not defined.
+		if ( window.mw ) {
```
mooeypoo: Alternatively, we could make this a class variable. In tests we mock it, and in instantiation we pass mw.Api through. That will keep the class a bit cleaner? Not a must, though.

src/App.js
```diff
@@ -70,35 +73,47 @@ class App {
...
+				const ids = this.api.getIdsFromElement( e.target );
+				// Activate all this user's contribution spans.
+				$( '.token-editor-' + ids.editorId ).addClass( 'active' );
+			} )
+			.on( 'mouseleave', () => {
+				// Deactivate all spans.
+				$( '.mw-parser-output .editor-token' ).removeClass( 'active' );
+			} );
+
+		// Create RevisionPopupWidget instance.
```
mooeypoo: Should this be inside the "attachContentListeners" ? I think the creation of the revision popup should be either in the constructor (saved as class var) or in start(), and you can call it here with `this.revisionPopup` (or even "just" `this.popup` since it's probably the only popup in the system (welcome popup is a separate entry/operation code anyways)

In any case, the creation of the popup is unrelated to attaching listeners really, and having it be a callable class variable is more architecturally consistent and can actually be useful later.

Example: If we ever have the need to hide the popup programmatically (like, we might need to listen to VE's activation hook, for example) then having it as a class variable is useful.

src/App.js
```diff
@@ -70,35 +73,47 @@ class App {
...
+		const revisionPopup = new RevisionPopupWidget();
+		$( document.body ).append( revisionPopup.$element );
```
mooeypoo: why not just `$( 'body' )` ?

src/App.js
```diff
@@ -70,35 +73,47 @@ class App {
...
+
+		$( '.editor-token' ).on( 'click', e => {
+			const ids = this.api.getIdsFromElement( e.target );
+			revisionPopup.setContent( this.api.getTokenInfo( ids.tokenId ) );
```
mooeypoo: Arguably, you can group all of these into a single method in RevisionPopup... for example, something like this

```
RevisionPopup.prototype.processToken = function ( tokenInfo, $target ) {
     this.setContent( tokenInfo );
     this.setFloatableContainer( $target );
};
```

I can see pros and cons about having `toggle( true)` in or out of that method, but having it as a central method can centralize the general logic in case we want to add behavior or change it. Sounds a bit better architecturally, even though it's not crucial.

src/RevisionPopupWidget.js
```diff
@@ -0,0 +1,43 @@
...
+/* Setup */
+OO.inheritClass( RevisionPopupWidget, OO.ui.PopupWidget );
+
+/**
+ * Update the contents of the revision popup.
+ * @param {Object} data As returned by Api.getTokenInfo().
+ */
+RevisionPopupWidget.prototype.setContent = function ( data ) {
+	// Note that the English namespaces will normalize to the wiki's local namespaces.
+	const userLinks = `
```
mooeypoo, line 29: I have to admit, I can't decide what I think about this :D It absolutely 100% might be just personal preference... but would it not be clearer/nicer to jQuerify all of this?

src/RevisionPopupWidget.js
```diff
@@ -0,0 +1,43 @@
...
+		<a target="_blank" href="${mw.util.getUrl( `User:${data.username}` )}">${data.username}</a>
+		${mw.msg( 'parentheses-start' )}<a target="_blank" href="${mw.util.getUrl( `User talk:${data.username}` )}">${mw.msg( 'talkpagelinktext' )}</a>
+		${mw.msg( 'pipe-separator' )}
+		<a target="_blank" href="${mw.util.getUrl( `Special:Contributions/${data.username}` )}">${mw.msg( 'contribslink' )}</a>${mw.msg( 'parentheses-end' )}
+	`,
+		// eslint-disable-next-line no-undef
```
mooeypoo: Is it complaining about "moment" ?  If so, add it to `.eslintrc` as `"moment": false` in globals

src/RevisionPopupWidget.js
```diff
@@ -0,0 +1,43 @@
...
+		dateStr = moment( data.revisionTime ).locale( mw.config.get( 'wgUserLanguage' ) ).format( 'LLL' ),
+		diffLink = `<a target="_blank" href="${mw.util.getUrl( `Special:Diff/${data.revisionId}` )}">${dateStr}</a>`,
+		addedMsg = mw.message( 'ext-whowrotethat-revision-added', userLinks, diffLink ).parse(),
+		attributionMsg = mw.message( 'ext-whowrotethat-revision-attribution', data.score ).parse();
+
+	$( '.ext-wwt-revisionPopupWidget-content' ).html( `${addedMsg.trim()} ${attributionMsg}` );
```
mooeypoo: I know that we're the ones setting this data, but there still might be some unpredictable input; I'd use `$.parseHTML()` here for safety.

src/less/general.less
```diff
@@ -1,5 +1,13 @@
-.mw-parser-output .editor-token.active {
-	background-color: #facc48;
+.mw-parser-output .editor-token {
+	cursor: pointer;
+
+	a {
```
mooeypoo, line 4: We actually might want to do this for all links when the system's active... is that covering them all or just tokenized ones?

This is probably not for this task, but I wonder if we can (or should?) disable the hovercards behavior. Should we? I'll ping Ilana for that question.

mooeypoo (changes requested):
This overall looks SWEEEEEET.

Didn't have time to test it yet, that's going to be my next step. I wrote a few comments about architectural considerations but they're mostly pretty minor. It's really taking on a great shape!

src/Api.js
MusikAnimal: I was just going off of:
https://github.com/wikimedia/WhoWroteThat/blob/25a46b57cb0346e936348aaf322d35c84782aacc/src/ActivationSingleton.js#L85-L89

I don't think there's much benefit to testing the revision popup, as it is largely logicless, unless we wanted to assert the markup was generated properly.

src/App.js
MusikAnimal: Sounds good to me! My thinking was we didn't want to create the popup until WWT actually finished loading, but I can at least move it to a separate method that doesn't have to do with adding listeners.

src/App.js
MusikAnimal: I got that from you! ;-) https://app.slack.com/client/T024KLHS4/CDG6XHT3Q

`'body'` works just as well, I shall change it.

src/RevisionPopupWidget.js
MusikAnimal: Hehe, I almost asked you ahead of time. I started with jQuery then stopped; there's just a lot of markup and attributes. It would make this 4-line template into a really big method (assuming you want `.prop()` etc. to be on its own line). To me it's easier to read the raw markup as a template, similar to Twig. If we were all ES5 then I'd certainly be more inclined to use jQuery :)

I did consider however using an array of templates and joining them, which might make it even easier to read (note that we can't have whitespace between the parentheses and surrounding content):
```
const userLinks = [
    `<a target="_blank" href="${mw.util.getUrl( `User:${data.username}` )}">${data.username}</a>`,
    mw.msg( 'parentheses-start' ),
    `<a target="_blank" href="${mw.util.getUrl( `User talk:${data.username}` )}">${mw.msg( 'talkpagelinktext' )}</a>`,
    ...
].join();
```

src/less/general.less
MusikAnimal: The `pointer-events: none` will disable hovercards, but I don't think they'd work anyway because WWT replaces the content and the hovercard listeners aren't re-added.

This does only target tokenized links. I do think any non-tokenized area (like infoboxes) should be completely un-clickable, and possibly grayed out (as the greasemonkey version does).

mooeypoo:
Comment from testing: 

When the popup is appearing, we should freeze the highlight behavior, otherwise it creates a really weird and jarring view where highlight is changing to another piece but the popup remains where it is. 

This can be done inside the mouseover event if we "just" check whether the popup is visible or not (pretty sure there's a `isVisible()` getter on OO.ui.PopupWidget)

src/Api.js
mooeypoo: Aye, in that case, we needed the mw.messages.set thing to run, but didn't mind that it wouldn't be available in the test.

In this specific case, we could make the architectural case that the API class could "live" with different implementations of the API (just a quick example; mw.Api or mw.ForeignApi) and that *it* just expects a general type of API, but doesn't say what it is, and wherever we instantiate it, we give it the right class.

It's not really a must, I was just thinking about making the class less cluttered with dependencies, but it's not really crucial. Also, I believe that you're not using it in this specific PR anyways so it can just be taken out and we can think about the best way to add something like that in for the other ticket about the edit summary.

src/App.js
mooeypoo: hah!

I am actually not 100% sure if there's a good difference. I'm just more used to (and I think we use it elsewhere in WWT) to the `$('body')` version. I should look into the difference.

src/Api.js
MusikAnimal: No problem. I've changed it to accept an mw.Api instace. We are using it here in `fetchMessages()`

MusikAnimal:
> When the popup is appearing, we should freeze the highlight behavior, otherwise it creates a really weird and jarring view where highlight is changing to another piece but the popup remains where it is.

Good idea! Done.

samwilson (changes requested):
Looks great!

The only bug I could noticed was that when a popup is opened it's possible to click on other text and open another popup, but the hover highlighting just disappears (instead of moving to the new element's).

MusikAnimal:
> The only bug I could noticed was that when a popup is opened it's possible to click on other text and open another popup, but the hover highlighting just disappears (instead of moving to the new element's).

Fixed!

PR #42, opened 2019-09-01
Rewrote and added a couple of things.

Greek:
In install.md, the user can refer to the README file if they wanna
install WWT as a browser extension. In README.md, the build command
has been shortened and extra instructions have been added for gadgets

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ There's a Grunt job to output the code into a working browser extension. To test
 
 1. Clone the repo
 2. Run `npm install`
-3. Run `npm build`
+3. Run `npm run build`
 4. Testing in Chrome: Go to `chrome://extensions/`, click on 'Load unpacked', and choose the `WhoWroteThat/dist/extension` directory. Enable the extension, and go to any article on en.wikipedia.org.
 5. Testing in Firefox: [Follow the official instructions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Temporary_Installation_in_Firefox) to load `WhoWroteThat/dist/extension` directory as an unpacked addon. Enable, and go to any article on en.wikipedia.org. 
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -7,23 +7,29 @@ A browser extension (Chrome and Firefox) based on the https://github.com/wikiwho
...
 * Spanish Wikipedia: `*://es.wikipedia.org/*`
 
 ## Testing the Browser Extension
+
 There's a Grunt job to output the code into a working browser extension. To test it:
 
 1. Clone the repo
 2. Run `npm install`
-3. Run `./node_modules/.bin/grunt build`
+3. Run `npm build`
```
MusikAnimal: Should be `npm run-script build` or `npm run build`. This may be dependent on the version of npm, not sure.

tutorials/install.md
```diff
@@ -17,4 +17,4 @@ Installing WhoWroteThat
 
 ## As a browser extension
 
-1. Install it. (Details t.b.c.)
+Refer to [the README file](../README.md).
```
MusikAnimal: Eventually this would be a link to entry on the browser's extension marketplace (Google Web Store, etc.), not how to build it from source.

MusikAnimal (changes requested):
Hey, thanks for this! I think there may be some confusion here; it's my understanding that install.md is user-facing documentation on how to install WWT as a gadget or browser extension, that shouldn't involve the build process at all. README.md however has instructions for contributors to the codebase.

Things have changed since install.md was first written; you'll notice the JS and CSS links are broken because we're no longer committing the compiled assets. I think we're still figuring out how we want to expose these files without requiring interested parties build them from source.

As an aside, I think this could have a more description commit summary. "_Rewrote and added a couple of things_" is ambiguous. I thought it involved code changes. Maybe "_Improve installation instructions in README.md_"?

tutorials/install.md
Greek: Reverted this for now, since this probably won't be released soon on the web store.

README.md
Greek: Done.

MusikAnimal (approved):
I think typically if you wanted to test the on-wiki version, you'd just add the script in your userspace rather than create a gadget. But this change to the instructions seems fine for now, at least until we have a proper public-facing build system to copy/paste the compiled assets.

PR #43, opened 2019-09-04
Fix to make the click event use the correct target for thumbnails

MusikAnimal:
Thumbnail images are tokenized but are usually floated to the left or
right, meaning the element that triggered the event may differ from the
visible element that triggered the event (the one with the token ID).
Additionally, we need to attach the popup to the thumbnail itself.

Bug: T231959

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -128,7 +128,7 @@ class App {
 				if ( this.revisionPopup.isVisible() ) {
 					return;
 				}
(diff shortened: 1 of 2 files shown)


samwilson (approved):
Sorry, I thought I'd reviewed this last week. Looks good to me. Just needs to be rebased I think.

MusikAnimal:
Rebased! Merging based on Sam's approval.

PR #44, opened 2019-09-04
Fix display and links for IPs

MusikAnimal:
WikiWho prefixes IP contributions with "0|". We check for this and make
sure both the display of the username (IP) and the links are correct.

Bug: T231960


samwilson (approved):
Works in my testing.

PR #45, opened 2019-09-05
Add edit summary to revision details pop-up

MusikAnimal:
Bug: T228941

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -34,6 +34,34 @@ class Api {
 	}
 
 	/**
+	 * Get parsed edit summary for the given revision.
+	 * @param {number} revId
+	 * @return {Promise} Resolving Object with keys 'comment' and 'size'.
+ 	 */
+	fetchEditSummary( revId ) {
+		return this.mwApi.ajax( {
+			action: 'compare',
+			fromrev: revId,
+			torelative: 'prev',
+			prop: 'parsedcomment|size',
+			formatversion: 2
+		} ).then(
+			data => {
+				if ( data.compare ) {
+					return {
+						comment: data.compare.toparsedcomment,
+						size: data.compare.tosize - ( data.compare.fromsize || 0 )
+					};
+				}
+				return $.Deferred().reject();
+			},
+			failData => {
+				return $.Deferred().reject( failData );
+			}
+		);
+	}
+
+	/**
 	 * Get the value of a parameter from the given URL query string.
 	 *
 	 * @protected
--- a/src/less/RevisionPopupWidget.less
+++ b/src/less/RevisionPopupWidget.less
@@ -1,0 +1,7 @@
+.ext-wwt-revisionPopupWidget-attribution {
+	margin-top: 14px;
+}
+
+.mw-diff-bytes {
+	font-style: italic;
+}
--- a/src/ActivationSingleton.js
+++ b/src/ActivationSingleton.js
@@ -121,6 +121,8 @@ class ActivationSingleton {
 				'oojs-ui',
 				'oojs-ui.styles.icons-user',
 				'oojs-ui.styles.icons-interactions',
+				'mediawiki.interface.helpers.styles',
+				'mediawiki.special.changeslist',
 				'mediawiki.widgets.datetime',
 				'moment'
 			] )
--- a/src/App.js
+++ b/src/App.js
@@ -126,10 +126,21 @@ class App {
 			} );
 
 		$( '.editor-token' ).on( 'click', e => {
-			const ids = this.api.getIdsFromElement( e.target );
+			const ids = this.api.getIdsFromElement( e.target ),
+				tokenInfo = this.api.getTokenInfo( ids.tokenId );
 			this.activateSpans( ids.editorId );
-			this.revisionPopup.show( this.api.getTokenInfo( ids.tokenId ), $( e.target ) );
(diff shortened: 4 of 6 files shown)


MusikAnimal:
I did not make it conditionally show a colon `:` versus full stop `.` based on whether there's an edit summary. Certainly can, but I wasn't sure it was worth having an extra message for this tiny detail, along with the necessary logic. I've asked about this at https://phabricator.wikimedia.org/T228941#5466901

src/less/RevisionPopupWidget.less
```diff
@@ -0,0 +1,29 @@
+.ext-wwt-revisionPopupWidget-attribution {
+	margin-top: 14px;
+}
+
+/* This matches the native MediaWiki CSS for diff sizes, which may not be available on article pages. */
```
samwilson: I think if you add the `mediawiki.interface.helpers.styles` module then it'll work (and it's not a huge module). It'll also then use the `msg-brackets-start` and `msg-brackets-end` messages instead of hard-coded paretheses.

src/less/RevisionPopupWidget.less
```diff
@@ -0,0 +1,29 @@
...
+	&::before {
+		content: '(';
+	}
+
+	&::after {
+		content: ')';
+	}
+}
+
+.mw-plusminus-pos {
```
samwilson: I was going to say that these colours could be used from `mediawiki.special.changeslist`, but that's got other stuff in it that we don't really need. But maybe a comment saying where they're defined in MediaWiki would be good.

samwilson (commented):
Works great. Just some comments about the styles, but they're not blockers.

src/less/RevisionPopupWidget.less
```diff
@@ -0,0 +1,29 @@
...
+	margin-top: 14px;
+}
+
+/* This matches the native MediaWiki CSS for diff sizes, which may not be available on article pages. */
+
+.mw-diff-bytes {
+	font-style: italic;
+
+	&::before {
+		content: '(';
```
mooeypoo: Don't put parentheses hard coded like this; it's messing up RTL, and won't work for interface languages that have different definition for parentheses, like French.

MW has a `parentheses` message, use that?

samwilson: (See my comment above.) We can probably use the mediawiki.special.changeslist module.

samwilson: No, not that one, oops. mediawiki.interface.helpers.styles :)

src/less/RevisionPopupWidget.less
MusikAnimal: Nice! I had no idea.

src/less/RevisionPopupWidget.less
MusikAnimal: Just now saw this comment; I ended up using `mediawiki.special.changeslist`. It doesn't seem very big so hopefully okay?

src/Api.js
```diff
@@ -33,6 +33,31 @@ class Api {
...
+	 * @return {Promise} Resolving Object with keys 'comment' and 'size'.
+ 	 */
+	fetchEditSummary( revId ) {
+		return this.mwApi.ajax( {
+			action: 'compare',
+			fromrev: revId,
+			torelative: 'prev',
+			prop: 'parsedcomment|size',
+			formatversion: 2
+		} ).then( data => {
```
mooeypoo: Formattingwise, I think it'll be easier to read with a newline:
```
} ).then(
    data => { 
       ...
    },
    failData => {
      ...
    }
);
```

But, granted, it might just be a personal preference :)

src/Api.js
```diff
@@ -33,6 +33,31 @@ class Api {
...
+			if ( data.compare ) {
+				return $.Deferred().resolve( {
```
mooeypoo: You don't have to wrap this with a .resolve, because you're aleady in a resolved path (you do this when you want to flip the promise from resolve to reject and vise versa like you do below on line 55)

Just return the object here directly.

src/Api.js
```diff
@@ -33,6 +33,31 @@ class Api {
...
+					comment: data.compare.toparsedcomment,
+					size: data.compare.tosize - ( data.compare.fromsize || 0 )
+				} );
+			}
+			return $.Deferred().reject();
+		}, failData => {
+			return $.Deferred().reject( failData );
```
mooeypoo: Similarly, you don't need to return a rejected promise, because the promise is already rejecting at this point, just return failData directly.

src/Api.js
MusikAnimal: That does look prettier! I wonder if there's an eslint option for this

src/Api.js
MusikAnimal: Done!

src/Api.js
samwilson: This doesn't work for me, if it's just

```
failData => {
    return failData;
}
```

If nothing's being done with `failData` then the whole error handler can just be left out, and it works correctly.

samwilson (changes requested):
^ my comment above

src/Api.js
MusikAnimal: Indeed! I've changed it back to returning a rejected promise.

`failData` is not being used, incidentally. I figure that's up to the caller of `fetchEditSummary()` to decide.

PR #47, opened 2019-09-06
Add exit functionality

samwilson:
Make it possible to close WWT from the info bar while loading,
and from the sidebar link. The link-manipulating is contained
within ActivationSingleton because that's where it's being
created.

https://phabricator.wikimedia.org/T231359
Bug: T231359

### DIFF
--- a/i18n/en.json
+++ b/i18n/en.json
@@ -4,6 +4,7 @@
 	},
 	"ext-whowrotethat-activation-link": "Who Wrote That?",
 	"ext-whowrotethat-activation-link-tooltip": "Activate Who Wrote That?",
+	"ext-whowrotethat-deactivation-link": "Close Who Wrote That?",
 	"ext-whowrotethat-state-pending": "<strong><em>Who Wrote That?</em></strong> is loading. This might take a while.",
 	"ext-whowrotethat-state-error": "API Error: $1",
 	"ext-whowrotethat-error-refresh": "Please refresh the page or try again later.",
--- a/i18n/qqq.json
+++ b/i18n/qqq.json
@@ -2,8 +2,9 @@
 	"@metadata": {
 		"authors": []
 	},
-	"ext-whowrotethat-activation-link": "The text for the link that toggles WhoWroteThat mode",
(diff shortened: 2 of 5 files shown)


i18n/en.json
```diff
@@ -4,6 +4,8 @@
 	},
 	"ext-whowrotethat-activation-link": "Who Wrote That?",
 	"ext-whowrotethat-activation-link-tooltip": "Activate Who Wrote That?",
+	"ext-whowrotethat-deactivation-link": "Close Who Wrote That?",
```
mooeypoo, line 7: OK, totally not a merge stopper, but we should flag this for Ilana and @prtksxna  ... this looks so weird as a question ...

mooeypoo (approved):
Looks good, didn't yet have time to test.

I will say that this is *starting* to look like we should start thinking about a general state-store (model?) for this. Right now, the states are called from basically two places (on/off with either of the [x] or link) but soon we'll have conditional states where the link text will need to change and toggle on/off for different conditions (with VE's loading) 

At that stage, it might be time to simplify things and get the "on/off" state of the entire system in a single place that all widgets consistently listen to and change their states and messages.

i18n/en.json
samwilson: Yes, and 'Close Who Wrote That?' with a tooltip of 'Deactivate Who Wrote That?' seems redundant. Maybe we don't have to change the link text at all?

mooeypoo: I don't think there's need for a tooltip for a text-based link. The tooltip is needed when we have an icon or button that the screen reader needs help parsing...

samwilson: Yeah that makes sense. Do you mean we should remove the existing tooltip as well? Although I guess that might be sorted out in [T231612](https://phabricator.wikimedia.org/T231612). I'll remove the new tooltip here.

MusikAnimal (approved):

MusikAnimal:
Works great! I shall merge, with the assumption https://github.com/wikimedia/WhoWroteThat/pull/47#pullrequestreview-284660533 can/will be addressed in a different PR.

PR #48, opened 2019-09-09
Add tiny PHP script to proxy requests to the WikiWho API

MusikAnimal:
Bug: T231492

### DIFF
--- a/public_html/wikiWhoProxy.php
+++ b/public_html/wikiWhoProxy.php
@@ -1,0 +1,30 @@
+<?php
+
+header( 'Content-Type: application/json' );
+header( 'Access-Control-Allow-Origin: *' );
+
+// Restrict access to Wikipedias.
+if ( 1 !== preg_match( '/^https:\/\/.*?\.wikipedia\.org/', $_SERVER[ 'HTTP_REFERER' ] ) ) {
+    http_response_code( 403 );
+    die();
+}
+
+$configPath = dirname( __FILE__, 2 ) . '/wikiwho.ini';
+$cnf = parse_ini_file( $configPath );
+if ( !$cnf || !$cnf[ 'user' ] || !$cnf[ 'password' ] ) {
+    throw new Exception(
+        'WikiWho API credentials not found at ' . $configPath
+    );
+}
+
+// Strip out /wikiwho for the Toolforge location.
+$endpoint = preg_replace( '/^\/wikiwho/', '', $_SERVER['REQUEST_URI'] );
+$redirectUrl = "https://api.wikiwho.net$endpoint";
+$context = stream_context_create( [
+    'http' => [
+        'method' => 'GET',
+        'header' => 'Authorization: Basic ' . base64_encode( $cnf[ 'user' ] . ':' . $cnf[ 'password' ] ),
+    ],
+] );
+
+echo file_get_contents( $redirectUrl, false, $context );
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ temp/
 dist/
 .idea/
 docs/
+wikiwho.ini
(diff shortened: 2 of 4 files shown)


MusikAnimal:
I'm going to change WWT to use the proxy in a separate PR. I will first contact Fabian about getting the necessary quota.

MusikAnimal:
Also, the proxy is live on Toolforge right now. Try making requests from Wikipedia, e.g. https://tools.wmflabs.org/wikiwho/en/whocolor/v1.0.0-beta/Hanksy/

public_html/wikiWhoProxy.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+header( 'Content-Type: application/json' );
+header( 'Access-Control-Allow-Origin: *' );
```
MusikAnimal, line 4: Or maybe we should just restrict to Wikipedias here, rather than checking the referrer? Some conditions such as altered security settings may bypass this, while the referrer would have to be spoofed (which is also pretty easy to do...)

public_html/wikiWhoProxy.php
```diff
@@ -0,0 +1,33 @@
...
+// Restrict access to Wikipedias.
+if ( 1 !== preg_match( '/^https:\/\/.*?\.wikipedia\.org/', $_SERVER[ 'HTTP_REFERER' ] ) ) {
+    http_response_code( 403 );
+    die();
+}
+
+$cnf = parse_ini_file( '../wikiwho.ini' );
+if ( !$cnf || !$cnf[ 'user' ] || !$cnf[ 'password' ] ) {
+    throw new Exception(
+        'WikiWho API credentials not found at ' . dirname( __FILE__, 2 ) . '/wikiwho.ini'
```
samwilson: The two different ways of constructing the path to wikiwho.ini confused me. Maybe a var?

public_html/wikiWhoProxy.php
```diff
@@ -0,0 +1,33 @@
...
+    );
+}
+
+// We strip out /wikiwho for the Toolforge location, and /wikiWhoProxy.php as the entry point when testing locally.
```
samwilson: How are you testing locally? Adding the Referer manually?

samwilson (commented):
Looks good to me.

Should this patch also handle some sort of switching of the URL in `src/config.js`? i.e. how do we actually tell the tool which API to use?

public_html/wikiWhoProxy.php
MusikAnimal: That was from before I added the referer check. I'll remove that part of the comment.

MusikAnimal:
> Should this patch also handle some sort of switching of the URL in src/config.js? i.e. how do we actually tell the tool which API to use?

We'll always use the Toolforge proxy, I think. I was going to make that change in a separate PR after we get the increased (likely unlimited) quota from Fabian.

samwilson:
> We'll always use the Toolforge proxy, I think.

Good point. I was confusing myself when thinking about how local development works, but of course we test from *.wikipedia.org anyway, so it doesn't matter.

> I was going to make that change in a separate PR after we get the increased (likely unlimited) quota from Fabian.

Sure.

public_html/wikiWhoProxy.php
samwilson: It doesn't seem very likely that anyone will abuse this, so I think how you've got it is fine. That said, we'll probably not even know if someone's hitting this too much unless we add some log-checking or something.

public_html/wikiWhoProxy.php
MusikAnimal: Done!

MaxSem (approved):

PR #49, opened 2019-09-11
Improve same-page linking

samwilson:
* Make sure section/citation link targets aren't obscured by the
  WhoWroteThat infobar, by adjusting scroll by the height of
  the bar.
* Add ext-wwt-enabled class to the parser-output element, so we
  can style links appropriately.
* Make all links in revision popups, including those in the edit
  summaries, open in new tabs.

Bug: T231595

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -102,6 +102,8 @@ class App {
 					this.widget.setErrorMessage( errorCode );
 				}
 			);
+		// Add a class for CSS to style links appropriately. Is removed in this.onWidgetClose().
+		activationInstance.getContentWrapper().addClass( 'ext-wwt-enabled' );
 	}
 
 	/**
@@ -120,7 +122,10 @@ class App {
 	}
 
 	/**
-	 * Add listener to highlight attribution and show the RevisionPopupWidget.
+	 * Add listeners to:
+	 *   - highlight attribution;
+	 *   - show the RevisionPopupWidget; and
+	 *   - scroll to the right place for fragment links.
 	 */
 	attachContentListeners() {
 		$( '.mw-parser-output .editor-token' )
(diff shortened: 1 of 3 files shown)


prtksxna:
Thanks for adjusting the scroll issue, @samwilson 😊

samwilson:
> Thanks for adjusting the scroll issue, @samwilson blush

No worries :) Do you think it should add a little bit more space than just `outerHeight( true )`? Because there's a drop shadow of 6px. (I briefly tried to retrieve that value in JS but wasn't sure it was necessary; otherwise, it'd just be hardcoded.)

MusikAnimal (changes requested):
Lookin' good! I think maybe the `pointer-events: none` thing isn't working for all links. For instance links in templates/infoboxes still work for me; e.g. the link in "_Media related to [Street artists] at Wikimedia Commons_" at https://en.wikipedia.org/wiki/Street_artist#External_links, or infobox at https://en.wikipedia.org/wiki/Angola_national_under-17_football_team. Same with category links.

Fragment links in edit summaries (such as section links) still open in a new tab. That seems fine to me, just wanted to point it out, because other fragment links to the same page don't open in a new tab.

Also, if you click on fragment links like in the table of contents, the URL isn't updated as it would normally. Not sure if this matters, either.

MusikAnimal (approved):
Never mind! I forgot that when there are CSS changes, I need to go back to chrome://extensions and refresh the extension. `pointer-events: none` is working as it should :)

The notes about fragment links still stand, however.

samwilson:
> Fragment links in edit summaries (such as section links) still open in a new tab. That seems fine to me, just wanted to point it out, because other fragment links to the same page don't open in a new tab.

Yes, this is by design, but I agree with you that it feels slightly odd. It makes sense for other links in edit summaries, and section links may be referring to sections that no longer exist (so opening in the new tab will just leave them at the top of the page anyway). I wondered about making them jump to the right section in that revision, because it'll always exist there, but that's not edit summary links work in other places (such as page history) so it seems inconsistent to do it here.

> Also, if you click on fragment links like in the table of contents, the URL isn't updated as it would normally. Not sure if this matters, either.

Good catch, thanks. Fixed by manually setting `location.hash` (this isn't the terrible hack it sounds like, honest).

src/App.js
```diff
@@ -155,6 +160,26 @@ class App {
...
+		 */
+		$( "a[href^='#']" ).on( 'click', event => {
+			var targetId, linkOffset, infobarHeight;
+			if ( !this.widget.isVisible() ) {
+				// Use the default if WWT is not active.
+				return;
+			}
+			event.preventDefault();
+			// After preventing the default event from doing it, set the URL bar fragment manually.
+			window.location.hash = targetId;
```
MusikAnimal, line 177: This is `undefined` in my testing. Shouldn't this go below where `targedId` is set?

MusikAnimal (changes requested):

src/App.js
samwilson: Oh, you want it to *work* as well? So picky! :-P

(Fixed. And I'll write "I must not rearrange code after testing" 100 times.)

MusikAnimal (approved):

PR #50, opened 2019-09-13
Add web-ext for linting and building the extension

samwilson:
Add new grunt tasks for linting and building the browser
extension. This builds the zip file (in dist/) that's
required for submitting to the Firefox Add-ons store.

This also switches from grunt-run to grunt-shell, to move
all of the build process into Gruntfile and out of
package.json.

The gecko extension ID is removed from the manifest.json
because an ID will be generated when the extension is signed,
and this part of that file has been raising a warning on
Chrome.

The web-ext lint Grunt task is not part of the normal lint
Grunt task because it only works after the extension is built
(whereas other linting is before build), and anyway we might
want to run (all other) linting on its own.

Bug: T232802

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -27,8 +27,9 @@ module.exports = function Gruntfile( grunt ) {
 		};
 
 	grunt.loadNpmTasks( 'grunt-browserify' );
-	grunt.loadNpmTasks( 'grunt-run' );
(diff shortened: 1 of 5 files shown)


samwilson:
This is ready for review. I'll add fixes for Firefox's code review comments in a separate patch after this is merged.

PR #51, opened 2019-09-16
Add username associated with selected text in the top area

cimurah:
Display the username associated with the hightlighted text in the top area. The username shows in
the standard naming format.

Bug: T226746

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -167,7 +167,7 @@ class Api {
 	 * that represents the token info or false if a token wasn't found.
 	 */
 	getTokenInfo( tokenId ) {
-		let revId, revision, username, score;
+		let revId, revision, username, isIP, score;
 
 		// Get the token information. results.tokens structure:
 		// [ [ conflict_score, str, o_rev_id, in, out, editor/class_name, age ], ... ]
@@ -194,10 +194,16 @@ class Api {
 			}
 		}
 
+		// The API returns an IP that starts with '0|' when the user account is an IP
+		isIP = username.slice( 0, 2 ) === '0|';
+		// Clean up username by removing the first 2 chars if username is IP
+		username = isIP ? username.slice( 2 ) : username;
+
 		// Put it all together.
 		return {
 			username: username,
 			userId: token[ 5 ],
+			isIP: isIP,
 			revisionId: revId,
 			revisionTime: new Date( revision[ 0 ] ),
 			score: score
--- a/src/InfoBarWidget.js
+++ b/src/InfoBarWidget.js
@@ -1,3 +1,5 @@
+import Tools from './Tools';
+
 /**
  * @class
  * @param {Object} config Configuration options.
@@ -17,6 +19,7 @@ const InfoBarWidget = function InfoBarWidget( config = {} ) {
 		flags: [ 'invert' ],
 		classes: [ 'ext-wwt-infoBarWidget-close' ]
 	} );
+	this.userInfoUsernameLabel = new OO.ui.LabelWidget();
 	this.userInfoLabel = new OO.ui.LabelWidget( {
 		label: mw.msg( 'ext-whowrotethat-ready-general' ),
 		classes: [ 'ext-wwt-infoBarWidget-info' ]
@@ -43,6 +46,7 @@ const InfoBarWidget = function InfoBarWidget( config = {} ) {
 			this.$pendingAnimation,
 			this.$icon,
 			this.$label,
+			this.userInfoUsernameLabel.$element,
 			this.userInfoLabel.$element,
 			this.closeIcon.$element
 		);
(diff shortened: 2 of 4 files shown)


i18n/en.json
```diff
@@ -17,5 +17,6 @@
 	"ext-whowrotethat-tour-welcome-description": "Activate the the tool from the sidebar. Hover over the text to see the author and their other contributions to this page. Click to get more details.",
 	"ext-whowrotethat-tour-welcome-dismiss": "Got it!",
 	"ext-whowrotethat-revision-added": "$1 added this on $2.",
-	"ext-whowrotethat-revision-attribution": "They have written <strong>$1%</strong> of the page."
+	"ext-whowrotethat-revision-attribution": "They have written <strong>$1%</strong> of the page.",
+	"ext-whowrotethat-revision-username": "User:$1"
```
mooeypoo: Hmm. `User:` can be gendered in some languages (like Hebrew, and I think Spanish?). Most cases, we'd solve it by adding a {{GENDER}} rule to the translation *but* that rule takes into account the viewer, not the user that's presented. To show the user presented, we'd need to ask the server per user coming in.

Seeing as that's the case, honestly, we should talk to Product and maybe just display the username without the `User:` prefix.

For the moment, I think it's okay to leave that in (I won't consider it a blocker for merging) but it's definitely something we should consider removing or dealing with, as it will look very odd and not consistent with the rest of the wiki in gendered languages.

i18n/qqq.json
```diff
@@ -17,5 +17,6 @@
 	"ext-whowrotethat-tour-welcome-description": "The description for the welcome popup that appears when the extension is enabled for the first time, explaining the activation and base usage of the extension.",
 	"ext-whowrotethat-tour-welcome-dismiss": "Button to permanently dismiss the welcome popup that appears when the extension is enabled for the first time.",
 	"ext-whowrotethat-revision-added": "Message shown in a popup that indicates who authored the text of an article. $1 is the username along with (talk | contribs) links, and $2 is a link to the diff with the timestamp as the link text.",
-	"ext-whowrotethat-revision-attribution": "Message indicating how much of the page an editor authored. This is shown immediately after the {{wm-msg|ext-whowrotethat-revision-added}} message."
+	"ext-whowrotethat-revision-attribution": "Message indicating how much of the page an editor authored. This is shown immediately after the {{wm-msg|ext-whowrotethat-revision-added}} message.",
+	"ext-whowrotethat-revision-username": "The text that displays in the info bar that indicates who authored the text of an article."
```
mooeypoo: You should document the parameter given here; We usually add `Parameters:\n* $1 - [explanation]` at the end of the qqq message. You can see `ext-whowrotethat-error-contact` message for an example

src/Api.js
```diff
@@ -183,6 +183,8 @@ class Api {
 		revId = token[ 2 ];
 		revision = this.results.revisions[ revId ];
 		username = revision[ 3 ];
+		isIP = username.slice( 0, 2 ) === '0|';
```
mooeypoo: What's the format that we're getting from the API that starts with this? Can you quickly add a comment above? it looks weird to have a `'0|` as an indication of an IP address unless it's how the API specifically gives it. Does the API give us `username|detail` and then username is 0 if it's not logged in?

Just add a quick comment right above that to clarify for the future.

src/InfoBarWidget.js
```diff
@@ -118,4 +120,18 @@ InfoBarWidget.prototype.setErrorMessage = function ( errCode = 'refresh' ) {
...
+ *
+ * @param {string} username
+ */
+InfoBarWidget.prototype.setUsernameInfo = function ( username ) {
+	this.userInfoUsernameLabel.setLabel( mw.msg( 'ext-whowrotethat-revision-username', username ) );
+};
+
+// Clear the username information
+InfoBarWidget.prototype.clearUsernameInfo = function () {
+	this.userInfoUsernameLabel.setLabel( '' );
```
mooeypoo, line 136: We should look into the usability here; when the screen is narrower, and the username is longer, the info bar jumps to accommodate 2 lines instead of one. Flagging this as something we should check into when QA'ing and for product to decide.

mooeypoo (changes requested):
Overall looks great, but we might want to ask for some product followups later.
Minor requests for documentation inline.

i18n/en.json
mooeypoo: Spoke and approved with product -- do not show `User:` prefix at all.

src/InfoBarWidget.js
mooeypoo: I showed this to Ilana; don't bother touching this in this PR, but it's a good-to-know, especially when we deal with languages like German where the sentences can be a lot longer.

We'll keep an eye on this in QA and see if we need to do some UX followup.

src/Api.js
MusikAnimal: Yup, this is how WikiWho differentiates accounts from IPs. This code was refactored from https://github.com/wikimedia/WhoWroteThat/blob/997016b08ed367e0a9bd47865bcb5989b8de99ee/src/RevisionPopupWidget.js#L46-L47 I should have added a comment there.

src/InfoBarWidget.js
cimurah: Sounds good, I did notice it. We can add a placeholder to avoid the jump. We'll see what product decides :)

MusikAnimal (approved):
Works great, though there is the outstanding issue that some usernames/IPv6 addresses could cause the infobar to jump around, etc. It seems we're okay with following up with that in a separate PR, as necessary, and Moriel's other concerns appear to have been addressed, so I'm merging! Also I really need this code for #54 :)

PR #52, opened 2019-09-17
Retry API requests on 'refresh' errors

samwilson:
If a WikiWho request fails with a retry-able error, retry it up
to 3 times. No UI feedback is given, but progress is output to
the console.

https://phabricator.wikimedia.org/T232232

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -212,37 +212,57 @@ class Api {
 	 * or rejects if there was an error.
 	 */
 	getData( wikiUrl ) {
-		const api = this;
+		let getJsonData,
+			retry = 1,
+			retries = 4;
 		if ( this.resultsPromise ) {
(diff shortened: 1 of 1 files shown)


MusikAnimal (approved):
I tested this by loading WikiWho on a brand new article (just choosing from the top of Special:NewPagesFeed). In my case there genuinely wasn't any data for that article yet, but after the 3rd retry, it loaded! Very slick.

My only suggestion might be to continually increase the delay before retrying, in the event say this is a very large article, and it takes that much longer for WikiWho to generate the data. So instead of using `retryWait`, we could use `1000 * retries`. Just a thought.

samwilson:
Good idea! Done.

MusikAnimal (approved):
Works wonderfully!

PR #53, opened 2019-09-19
Clone original content later

samwilson:
Add a setter to ActivationSingleton so the $originalContent can
be stored when WWT is opened, so other gadgets etc. have already
modified the content.

Bug: T231424


MusikAnimal (approved):
Works! You no longer can click on images (following T231595), so I tested this by loading WWT, exiting out, and ensuring the links for the images still go to Commons. Works like a charm.

PR #54, opened 2019-09-19
Implement loading state support for edit summary

MusikAnimal:
Minor refactoring of RevisionPopupWidget.

Fix to Tools.bidiIsolate() to permit HTML.

Bug: T231115

### DIFF
--- a/src/RevisionPopupWidget.js
+++ b/src/RevisionPopupWidget.js
@@ -22,6 +22,11 @@ const RevisionPopupWidget = function RevisionPopupWidget() {
 /* Setup */
 OO.inheritClass( RevisionPopupWidget, OO.ui.PopupWidget );
 
+/**
+ * Get markup for the diff size.
+ * @param {number} size
+ * @return {string}
+ */
 function getSizeHtml( size ) {
 	let sizeClass;
 
@@ -37,6 +42,31 @@ function getSizeHtml( size ) {
 		<span class="${sizeClass} mw-diff-bytes">` +
 			`${size > 0 ? '+' : ''}${mw.language.convertNumber( size )}` +
 		'</span>';
+}
+
+/**
+ * Get markup for the edit summary.
+ * @param {Object} data As returned by Api.prototype.getTokenInfo().
+ * @return {string}
+ */
+function getCommentHtml( data ) {
+	if ( data.comment === '' ) {
+		// No edit summary.
+		return '';
+	} else if ( data.comment === undefined ) {
+		// Not yet available.
+		return `
+			<div class="ext-wwt-revisionPopupWidget-comment">
+				<div class="ext-wwt-shimmer ext-www-shimmer-animation"></div>
+				<div class="ext-wwt-shimmer ext-www-shimmer-animation"></div>
+			</div>`;
+	}
+
+	return `
+		<div class="ext-wwt-revisionPopupWidget-comment ext-wwt-revisionPopupWidget-comment-transparent">
+			<span class="comment comment--without-parentheses ext-wwt-revisionPopupWidget-comment">${Tools.bidiIsolate( data.comment, true )}</span>
+			${getSizeHtml( data.size )}
+		</div>`;
 }
 
 /**
@@ -83,7 +113,7 @@ function getUserLinksHtml( data ) {
 /**
  * Show the revision popup based on the given token data, above the given element.
  * Note that the English namespaces will normalize to the wiki's local namespaces.
- * @param {Object} data As returned by Api.getTokenInfo().
+ * @param {Object} data As returned by Api.prototype.getTokenInfo().
  * @param {jQuery} $target Element the popup should be attached to.
  */
 RevisionPopupWidget.prototype.show = function ( data, $target ) {
@@ -94,12 +124,12 @@ RevisionPopupWidget.prototype.show = function ( data, $target ) {
 			.attr( 'href', mw.util.getUrl( `Special:Diff/${data.revisionId}` ) )
 			.text( dateStr ),
 		addedMsg = mw.message( 'ext-whowrotethat-revision-added', $userLinks, $diffLink ).parse(),
-		commentMsg = data.comment ?
-			`<span class="comment comment--without-parentheses ext-wwt-revisionPopupWidget-comment">${Tools.bidiIsolate( $.parseHTML( data.comment ), true )}</span>` :
-			'',
-		sizeMsg = data.size ? getSizeHtml( data.size ) : '',
 		attributionMsg = `<div class="ext-wwt-revisionPopupWidget-attribution">${mw.message( 'ext-whowrotethat-revision-attribution', data.score ).parse()}</div>`,
-		html = `${addedMsg.trim()} ${commentMsg}${sizeMsg} ${attributionMsg}`;
+		html = $.parseHTML( `
+			${addedMsg.trim()}
+			${getCommentHtml( data )}
+			${attributionMsg}
+		` );
 
 	this.$popupContent.html( html );
 
@@ -113,6 +143,11 @@ RevisionPopupWidget.prototype.show = function ( data, $target ) {
 
 	this.setFloatableContainer( $target );
 	this.toggle( true );
+
+	// Animate edit summary, if present.
+	if ( data.comment ) {
+		$( '.ext-wwt-revisionPopupWidget-comment' ).removeClass( 'ext-wwt-revisionPopupWidget-comment-transparent' );
+	}
 };
 
 export default RevisionPopupWidget;
--- a/src/Tools.js
+++ b/src/Tools.js
@@ -12,18 +12,19 @@ class Tools {
 	 * within a bidi isolation block.
 	 *
 	 * @param {jQuery|string} $content Content to isolate
-	 * @param {boolean} returnRawHtml Force the return value to
+	 * @param {boolean} [returnRawHtml] Force the return value to
 	 *  be raw html. If set to false, will return the encompassed
 	 *  jQuery object.
 	 * @return {jQuery|string} BiDi isolated jQuery object or HTML
 	 */
-	static bidiIsolate( $content, returnRawHtml ) {
+	static bidiIsolate( $content, returnRawHtml = false ) {
 		const $result = $( '<bdi>' );
+
 		if ( typeof $content === 'string' ) {
-			$result.text( $content );
-		} else {
-			$result.append( $content );
+			$content = $( $.parseHTML( $content ) );
 		}
+
+		$result.append( $content );
 
 		if ( returnRawHtml ) {
 			// `html()` sends the inner HTML, so we wrap the node
--- a/.eslintrc
+++ b/.eslintrc
@@ -17,6 +17,7 @@
         "sourceType": "module"
     },
     "rules": {
-        "arrow-parens": [ "error", "as-needed" ]
+        "arrow-parens": [ "error", "as-needed" ],
+        "require-jsdoc": [ "error" ]
     }
 }
--- a/src/Api.js
+++ b/src/Api.js
@@ -184,6 +184,10 @@ class Api {
 		revision = this.results.revisions[ revId ];
 		username = revision[ 3 ];
 
+		// WikiWho prefixes IP addresses with '0|'.
+		isIP = username.slice( 0, 2 ) === '0|';
+		username = isIP ? username.slice( 2 ) : username;
+
 		// Get the user's edit score (percentage of content edited).
 		// results.present_editors structure:
 		// [ [ username, user_id, score ], ... ]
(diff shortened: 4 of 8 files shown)


MusikAnimal:
@prtksxna This doesn't include the fade-in animation. This is because the implementation re-renders the markup for the popup, rather than manipulating it. We _could_ make it do the latter, but I think the re-render strategy offers cleaner code (presentation stays in RevisionPopupWidget.js, not where the edit summary fetching happens). I think it looks quite good without the fading (feels snappier, at least), but let me know what you think, and we'll adjust accordingly!

src/RevisionPopupWidget.js
```diff
@@ -36,6 +36,26 @@ function getSizeHtml( size ) {
 		'</span>';
 }
 
+function getCommentHtml( data ) {
```
mooeypoo, line 52: Please add documentation for this method

(Also, we missed it for the other method above, can be good to add there too)

mooeypoo (dismissed):
It's a tad weird, to be honest; in fast connections, it still jumps without enough time to see the blurry-boxed lines, and since it's not an instantly-recognizable pending animation, I don't know if it's not weirder to have it "jump" between that and the text.

That said, I tested it with a slowed down connection and it looks pretty awesome.... so... I'm not sure what the best approach is, other than perhaps going back to @prtksxna's original suggestion of delaying this animation by 100ms and only showing it if needed. 

We can do that as a followup if needed (Test and see if you like what you see first?) but we should pay attention to that.

src/RevisionPopupWidget.js
MusikAnimal: Doh! I am religious about this, but apparently just as forgetful, hehe. We should add https://eslint.org/docs/rules/require-jsdoc

prtksxna:
@MusikAnimal I don't think I completely follow what you're saying. Do you mean there is no fade-in for the edit summary (once it becomes available)?

@mooeypoo Yeah the jump might be pretty bad on faster connections. Another idea I had was to show the animation for a minimum ~250ms regardless of whether we have the data or not. This makes things a little slower for everyone but might be easier to implement compared to my original proposal.

MusikAnimal:
> @MusikAnimal I don't think I completely follow what you're saying. Do you mean there is no fade-in for the edit summary (once it becomes available)?

With this implementation, no. I was just worried about the separation of concerns thing, where RevisionPopupWidget handles presentation. Right now it was built to just be re-rendered as needed. I assumed the fade in/out stuff would need to happen in app.js, where we do our AJAX, but I suppose RevisionPopupWidget can safely make assumptions on what animations should be performed based on whether the edit summary is `null` (not available yet).

> Another idea I had was to show the animation for a minimum ~250ms regardless of whether we have the data or not. This makes things a little slower for everyone but might be easier to implement compared to my original proposal.

I'm going to give this a try. Same semi-problem as what I mentioned above in regards to keeping business logic out of RevisionPopupWidget, but I think I can still avoid this. I will try to add the fading, too.

MusikAnimal:
Okay! I've added in a minimum 250ms delay, and added a fade-in animation. There is no fade-out, but with some reworking I think we can implement it. The code is still relying on re-rendering the whole popup when the content changes. But overall though I think it looks pretty slick.

This PR borrows from #51 and #55. Maybe those should be merged first to preserve attribution.

@prtksxna Are you able to test out this PR? If not I can make a screencast.

src/Tools.js
```diff
@@ -12,15 +12,15 @@ class Tools {
...
 	 *  be raw html. If set to false, will return the encompassed
 	 *  jQuery object.
 	 * @return {jQuery|string} BiDi isolated jQuery object or HTML
 	 */
-	static bidiIsolate( $content, returnRawHtml ) {
+	static bidiIsolate( $content, returnRawHtml = false ) {
 		const $result = $( '<bdi>' );
 		if ( typeof $content === 'string' ) {
-			$result.text( $content );
+			$result.html( $content );
```
MusikAnimal: @mooeypoo I had to change this to leave HTML unescaped because the API sometimes gives us markup (for section or other internal links). The API escapes any other markup, so I think we're safe? Though I realize this function is probably meant to be used on anything, and not just edit summaries. Is there a better to do this?

mooeypoo: So the point of this is that the tool is expecting the user of the tool to do the safety check for it. If it's a string, it will be treated as such, on purpose. We do something similar in mw.msg internals.

This forces the use case that calls this function to intentionally encapsulate. So, the best way to do it is wrap the response in a jQuery object before sending it to this utility. 

It's mostly a conceptual way to make sure that when we read the code in x years, we see very clearly that whatever was (and wasn't) escaped was done purposefully, rather than trying to convince ourselves all over again if it's safe.

Does that make sense? I'd just encapsulate in jQuery before sending here. 

We do this in ooui `setLabel` too, for the same reasons.

MusikAnimal: Done! Except I did it in `Tools.bidiIsolate()`, which was already excepting raw HTML.

src/Tools.js
```diff
@@ -12,19 +12,20 @@ class Tools {
...
 	 *
 	 * @param {jQuery|string} $content Content to isolate
-	 * @param {boolean} returnRawHtml Force the return value to
+	 * @param {boolean} [returnRawHtml] Force the return value to
 	 *  be raw html. If set to false, will return the encompassed
 	 *  jQuery object.
 	 * @return {jQuery|string} BiDi isolated jQuery object or HTML
 	 */
-	static bidiIsolate( $content, returnRawHtml ) {
+	static bidiIsolate( $content, returnRawHtml = false ) {
```
mooeypoo, line 20: You don't actually need to specify that, it's falsey by default already.

src/Tools.js
```diff
@@ -12,19 +12,20 @@ class Tools {
...
 		const $result = $( '<bdi>' );
+
 		if ( typeof $content === 'string' ) {
-			$result.text( $content );
-		} else {
-			$result.append( $content );
+			$content = $( $.parseHTML( $content ) );
```
mooeypoo, line 24: So, I think we have different perspectives on the role of the function and the angle of where the future-proofing comes from.

Just as a general comment, `$.parseHTML()` is not a safety check, it just parses the string. We are only shifting the responsibility of verifying safety from the caller to this method.

With your approach (and please correct me if I've understood it wrong) you set things up where the method outputs the "correct" value all the time, no matter what the user of the method gives it. That means that we prioritize the ease of the user of the method (and make all uses of it simpler), but make the code more vague about how it protects data that may (perhaps even in the future usage of this method) not be as we expect.

With my approach, we intentionally put the responsibility of what is sent and how to handle it, specifically, on the user of the method. This means that I prioritize the generalization of the method at the "expense" of the user of the method, but it also means that the place where we decide how to handle a given data is the place where the data is in its context, which makes it a lot easier for us to note edge cases and to "convince ourselves" (and others who read the code in x years) which data provides what kind of functionality.

It also means that it exposes edge cases much quicker, which forces the user of the function to decide whether these uses deserve a different way of handling its data.

For example, let's say we collect a certain data element from some API request. Even if we assume that all API requests are safe (which is not necessarily always true, but true enough for now) we will send that API request as string to the wrapping method.

If we go with my approach, we will immediately see that there's something unique about that data that we may not have immediately expected or anticipated because all HTML would appear, immediately, unparsed. We'd look at it and know that we need to parse the HTML, that the data piece *may contain* html (or wikitext) and have a chance to decide how secure or insecure it is.

With your approach, the string will work out of the box -- which is convenient -- but it would also mean we risk missing those steps of being certain of every piece of data we're handling.

I don't think either one of us is wrong or right. I think it's a matter of priorities of where we see the different pieces of the code operates. For me personally, I prefer strategies that, while may look a tad bit more verbose at times, are also making the entire code immediately clear about what kind of data it handles, even if we look back at this code years and years from now, or even if we iterate on code and keep adding different features where snippets that are given to us are from various sources.

I'm okay going with your method (I prefer mine but I won't block merging over this preference ;) ) but I do think that if we *do* go with yours, we should compensate by having a comment written on the "user" side (the code user, where we call this method) that specifies whether we get HTML or plaintext, for posterity.

src/Tools.js
```diff
@@ -12,19 +12,20 @@ class Tools {
...
 	 * @return {jQuery|string} BiDi isolated jQuery object or HTML
 	 */
-	static bidiIsolate( $content, returnRawHtml ) {
+	static bidiIsolate( $content, returnRawHtml = false ) {
 		const $result = $( '<bdi>' );
+
 		if ( typeof $content === 'string' ) {
-			$result.text( $content );
-		} else {
-			$result.append( $content );
```
mooeypoo, line 25: You can also switch to doing
```
if ( typeof $content === 'string' ) {
   $result.html( $.parseHTML( $content ) );
} else {
   $result.append( $content );
}
```
Since `.html()` can accept a parsed html string and doesn't require a jquery object.

src/Tools.js
mooeypoo: Another piece that occurred to me now is that with my method, we can actually give more control to the caller; so, if you know that you have a string that might *seem* like html but you want it to not render, you intentionally send it in as a string. With the change you're proposing, there's basically no way for you to purposefully do that.

An example would be a username that's something like `&nbsp;`. Those may well be legal characters for a user but it's valid for any other string (I'm not sure what we're using in Wikipedias, but according to [$wgInvalidUsernameCharacters](https://www.mediawiki.org/wiki/Manual:$wgInvalidUsernameCharacters) the default illegal characters are just `@` and `:`.) So, if you have that user, you actually want it to appear as `&nbps;` and not a space. But if we take all strings and HTMLify them, we will never be able to effectively do that unless we force them through htmlentities or something similar. I think that might be the case with our username policy, though I'm not sure (marked it below for your consideration)  

Again, I think there's no real right or wrong here. I am okay changing things if you think it works better, and if we do encouter the above cases, we can edge case them specifically, too.

src/RevisionPopupWidget.js
```diff
@@ -39,49 +44,83 @@ function getSizeHtml( size ) {
...
+function getUserLinksHtml( data ) {
+	const contribsUrl = mw.util.getUrl( `Special:Contributions/${data.username}` ),
+		// We typically link to Special:Contribs for IPs.
+		userPageUrl = data.isIP ? contribsUrl : mw.util.getUrl( `User:${data.username}` );
+
+	return $( [] )
+		.add(
+			$( '<a>' )
+				.attr( 'href', userPageUrl )
+				.append( Tools.bidiIsolate( data.username ) )
```
mooeypoo: I wrote a lengthier comment about this below, but I just want to verify; this data.username will always be rendered as HTML even if it's purposefully not supposed to. Is that okay? I think we have most relevant characters illegal in MW to prevent issues with that, but I'm actually not sure.

mooeypoo (changes requested):
Overall great, and the transition's pretty neat!

I added an explanation of my perspective with the change to the Tool method, but I don't think it's a show-stopper as long as we pick a consistent one (if we do yours, let's make sure to comment on the callers)

src/Tools.js
MusikAnimal: I'm aware :) I just thought it made it more clear that we're looking for a boolean (I tend to do the same thing in other languages). It is totally unnecessary. I shall remove it.

src/Tools.js
MusikAnimal: You might be overthinking this! There is no "my way", I just needed it not escape the HTML. You told me to wrap it in a jQuery object, so I did. The function you wrote allowed you to pass in a string, so I kept that. However you want to do it is fine with me. What is your suggestion?

> An example would be a username that's something like `&nbsp;`. ... according to $wgInvalidUsernameCharacters the default illegal characters are just `@` and `:`

Usernames also are also bound by [$wgLegalTitleChars](https://www.mediawiki.org/wiki/Manual:$wgLegalTitleChars), since they can have user/user talk pages. In total `#<>[]|{}/@:` are prohibited. HTML entities (such as `&nbsp;`) are not allowed, either.

src/RevisionPopupWidget.js
MusikAnimal: I don't think it's possible to have a username that is interpreted by the browser as anything other than plaintext. We should be safe :) But we can escape this too, if you want.

mooeypoo: Nah, I wasn't sure, and while I remembered that we prevent the characters that involve illegal titles, I couldn't find anything on mw.org that explained whether we actually allow for something like `&nbsp;` or something. Thanks for confirming!

src/Tools.js
mooeypoo: Oops, I think I wasn't clear. I meant that the **caller** should wrap the content it sends to this method with a jquery method, not that the method itself should do it (For the reasons I put up above).

And I hope this didn't come off too abruptly -- I didn't mean to suggest we're batting heads; I meant more like "this is the approach I'm taking, while this is where you're coming from". Both are totally valid, but since there's no clear "this is wrong or right" I just wanted to explain why I prefer the caller to do the work.

That said, this is perfectly valid. Thank you for putting up with my attempts at explaining the jumbled thoughts in my head ;) 

I need to retest this and then it can be merged.

mooeypoo:
OK tested, and this looks awwwweeessoomme.

I have one comment about the behavior, but this can be done in a followup -- should we still do the animation if the popup is reopened (so, the data has already been loaded) ?

We can do something with checking the caching, but it can totally be a followup to the behavior, if we even think it's a behavior to fix. Just raising this as a general question.

mooeypoo (approved):

PR #55, opened 2019-09-21
BiDi and reinforcement: Add BiDi helper for usernames and content lang

mooeypoo:
Any content that appears in the popup that comes from the content language
should be bidirectionally-isolated. This means that if our interface
language is Hebrew or any other RTL language, the usernames and/or
diff summaries should still render in the original content language
which may well be LTR.

This patch does a couple of cleanups and fortifications for the
rendering of html pieces:
- Adds a Tools class with 'bidi' helper and tests. This can be then
  used in other places in the system where we display usernames.
- Security:
  - Make sure anything served to the bidi isolation is fortified
    and escaped, or otherwise a raw string.
  - Make sure all _blank links have rel=noopener
  - Use templates for all except links, so we can make sure that
    link attributes are escaped.

### DIFF
--- a/src/RevisionPopupWidget.js
+++ b/src/RevisionPopupWidget.js
@@ -1,8 +1,12 @@
+import Tools from './Tools';
+
 /**
  * @class
  * @constructor
  */
 const RevisionPopupWidget = function RevisionPopupWidget() {
+	this.$popupContent = $( '<div>' )
+		.addClass( 'ext-wwt-revisionPopupWidget-content' );
 	// Parent constructor
 	RevisionPopupWidget.parent.call( this, {
 		padded: true,
@@ -11,8 +15,7 @@ const RevisionPopupWidget = function RevisionPopupWidget() {
 		// FIXME: 'force-left' for RTL languages
 		align: 'force-right',
 		hideWhenOutOfView: false,
-		$content: $( '<div>' )
-			.addClass( 'ext-wwt-revisionPopupWidget-content' )
+		$content: this.$popupContent
 	} );
 };
 
@@ -48,29 +51,49 @@ RevisionPopupWidget.prototype.show = function ( data, $target ) {
 		contribsUrl = mw.util.getUrl( `Special:Contributions/${username}` ),
 		// We typically link to Special:Contribs for IPs.
 		userPageUrl = isIP ? contribsUrl : mw.util.getUrl( `User:${data.username}` ),
-		userLinks = `
-			<a href="${userPageUrl}">${username}</a>
-			${mw.msg( 'parentheses-start' )}<a href="${mw.util.getUrl( `User talk:${username}` )}">${mw.msg( 'talkpagelinktext' )}</a>
-			${mw.msg( 'pipe-separator' )}
-			<a href="${contribsUrl}">${mw.msg( 'contribslink' )}</a>${mw.msg( 'parentheses-end' )}
-		`,
+		// Create links using the jQuery objects so they're properly escaped
+		$userLinks = $( [] )
+			.add(
+				$( '<a>' )
+					.attr( 'href', userPageUrl )
+					.append( Tools.bidiIsolate( username ) )
+			)
+			.add( document.createTextNode( ' ' + mw.msg( 'parentheses-start' ) ) )
+			.add(
+				// Talk page
+				$( '<a>' )
+					.attr( 'href', mw.util.getUrl( `User talk:${username}` ) )
+					.text( mw.msg( 'talkpagelinktext' ) )
+			)
+			.add( document.createTextNode( ' ' + mw.msg( 'pipe-separator' ) + ' ' ) )
+			.add(
+				$( '<a>' )
+					.attr( 'href', contribsUrl )
+					.text( mw.msg( 'contribslink' ) )
+			)
+			.add( document.createTextNode( mw.msg( 'parentheses-end' ) ) ),
 		dateStr = moment( data.revisionTime ).locale( mw.config.get( 'wgUserLanguage' ) ).format( 'LLL' ),
-		diffLink = `<a href="${mw.util.getUrl( `Special:Diff/${data.revisionId}` )}">${dateStr}</a>`,
-		addedMsg = mw.message( 'ext-whowrotethat-revision-added', userLinks, diffLink ).parse(),
-		commentMsg = data.comment ? `<span class="comment comment--without-parentheses ext-wwt-revisionPopupWidget-comment">${data.comment}</span>` : '',
+		// Use jQuery to make sure attributes are properly escaped
+		$diffLink = $( '<a>' )
+			.attr( 'href', mw.util.getUrl( `Special:Diff/${data.revisionId}` ) )
+			.text( dateStr ),
+		addedMsg = mw.message( 'ext-whowrotethat-revision-added', $userLinks, $diffLink ).parse(),
+		commentMsg = data.comment ?
+			`<span class="comment comment--without-parentheses ext-wwt-revisionPopupWidget-comment">${Tools.bidiIsolate( $.parseHTML( data.comment ), true )}</span>` :
+			'',
 		sizeMsg = data.size ? getSizeHtml( data.size ) : '',
 		attributionMsg = `<div class="ext-wwt-revisionPopupWidget-attribution">${mw.message( 'ext-whowrotethat-revision-attribution', data.score ).parse()}</div>`,
-		$popupContent = $( '.ext-wwt-revisionPopupWidget-content' ),
-		html = $.parseHTML( `${addedMsg.trim()} ${commentMsg}${sizeMsg} ${attributionMsg}` );
+		html = `${addedMsg.trim()} ${commentMsg}${sizeMsg} ${attributionMsg}`;
 
-	$popupContent.html( html );
+	this.$popupContent.html( html );
 
 	if ( $target.find( '.thumb' ).length ) {
 		$target = $target.find( '.thumb' );
 	}
 
 	// Make sure all links in the popup (including in the edit summary) open in new tabs.
-	$popupContent.find( 'a' ).attr( 'target', '_blank' );
+	this.$popupContent.find( 'a' )
+		.attr( 'target', '_blank' );
 
 	this.setFloatableContainer( $target );
 	this.toggle( true );
--- a/src/Tools.js
+++ b/src/Tools.js
@@ -1,0 +1,36 @@
+const $ = require( 'jquery' );
+
+/**
+ * Class to hold some global helper tools
+ */
+class Tools {
+	/**
+	 * Encompass strings with BiDi isolation for RTL/LTR support.
+	 * This should be used on any string that is content language,
+	 * to make sure that if the user uses a different directionality
+	 * in the interface language, the output is shown properly
+	 * within a bidi isolation block.
+	 *
+	 * @param {jQuery|string} $content Content to isolate
+	 * @param {boolean} returnRawHtml Force the return value to
+	 *  be raw html. If set to false, will return the encompassed
+	 *  jQuery object.
+	 * @return {jQuery|string} BiDi isolated jQuery object or HTML
+	 */
+	static bidiIsolate( $content, returnRawHtml ) {
+		const $result = $( '<bdi>' );
+		if ( typeof $content === 'string' ) {
+			$result.text( $content );
+		} else {
+			$result.append( $content );
+		}
+
+		if ( returnRawHtml ) {
+			// `html()` sends the inner HTML, so we wrap the node
+			// and send the result to include the new <bdi> wrap
+			return $( '<div>' ).append( $result ).html();
+		}
+		return $result;
+	}
+}
+export default Tools;
--- a/test/suite/Tools.test.js
+++ b/test/suite/Tools.test.js
@@ -1,0 +1,43 @@
+import { expect } from 'chai';
+import Tools from '../../src/Tools';
+const $ = require( 'jquery' );
+
+describe( 'Tools test', () => {
+	describe( 'Tools.bidiIsolate', () => {
+		const cases = [
+			{
+				input: 'foo',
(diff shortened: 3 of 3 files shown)


src/RevisionPopupWidget.js
```diff
@@ -48,29 +51,52 @@ RevisionPopupWidget.prototype.show = function ( data, $target ) {
...
 		$target = $target.find( '.thumb' );
 	}
 
 	// Make sure all links in the popup (including in the edit summary) open in new tabs.
-	$popupContent.find( 'a' ).attr( 'target', '_blank' );
+	this.$popupContent.find( 'a' )
+		.attr( 'target', '_blank' )
+		// Fix security volunerability associated with target=_blank
+		// see example: https://developers.google.com/web/tools/lighthouse/audits/noopener
+		.attr( 'rel', 'noopener noreferrer' );
```
MusikAnimal: Interesting! If this is a real concern, our other tools (Event Metrics, XTools, Pageviews, etc.) are littered with the same vulnerability. For WWT these should be internal links though, right? Anchor tags are not generated for URLs in edit summaries. It seems like good practice nonetheless, but I would at least remove `noreferrer` since we probably have no reservations about this for internal links, and it helps analytics track where traffic came from.

I'm probably missing something, but I don't believe that `noreferrer` works all the time, because I almost always see a wiki URL as the referrer in the access logs for XTools/Pageviews.

src/RevisionPopupWidget.js
```diff
@@ -48,29 +51,52 @@ RevisionPopupWidget.prototype.show = function ( data, $target ) {
 		contribsUrl = mw.util.getUrl( `Special:Contributions/${username}` ),
 		// We typically link to Special:Contribs for IPs.
 		userPageUrl = isIP ? contribsUrl : mw.util.getUrl( `User:${data.username}` ),
-		userLinks = `
-			<a href="${userPageUrl}">${username}</a>
-			${mw.msg( 'parentheses-start' )}<a href="${mw.util.getUrl( `User talk:${username}` )}">${mw.msg( 'talkpagelinktext' )}</a>
-			${mw.msg( 'pipe-separator' )}
-			<a href="${contribsUrl}">${mw.msg( 'contribslink' )}</a>${mw.msg( 'parentheses-end' )}
-		`,
+		// Create links using the jQuery objects so they're properly escaped
```
MusikAnimal, line 54: Just curious, what would not be properly escaped? The template literals should respect any double-quotes in interpolated expressions, and we're using `$.parseHTML` on the final markup.

src/RevisionPopupWidget.js
```diff
@@ -48,29 +51,52 @@ RevisionPopupWidget.prototype.show = function ( data, $target ) {
...
-		diffLink = `<a href="${mw.util.getUrl( `Special:Diff/${data.revisionId}` )}">${dateStr}</a>`,
-		addedMsg = mw.message( 'ext-whowrotethat-revision-added', userLinks, diffLink ).parse(),
-		commentMsg = data.comment ? `<span class="comment comment--without-parentheses ext-wwt-revisionPopupWidget-comment">${data.comment}</span>` : '',
+		// Use jQuery to make sure attributes are properly escaped
+		$diffLink = $( '<a>' )
+			.attr( 'href', mw.util.getUrl( `Special:Diff/${data.revisionId}` ) )
+			.text( dateStr ),
+		addedMsg = mw.message( 'ext-whowrotethat-revision-added', $userLinks, $diffLink ).parse(),
+		commentMsg = data.comment ?
+			`<span class="comment comment--without-parentheses ext-wwt-revisionPopupWidget-comment">${Tools.bidiIsolate( $.parseHTML( data.comment ), true )}</span>` :
```
MusikAnimal, line 82: Do we need `$.parseHTML` here in addition to line 86?

src/RevisionPopupWidget.js
mooeypoo: Yup we might want to consider doing that in other places. This was all over twitter a couple of weeks ago, I don't think a lot of people know about it -- but it's pretty huge.

The recommendation is to add noreferrer but I think Firefox doesn't recognize it, so noopener is the alternative to get better support.

Here's an interesting article about the potential "attack" vector: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

I wouldn't say it's an urgent horrible thing to fix, but it might be a good reinforcement to add whenever possible.

src/RevisionPopupWidget.js
mooeypoo: So $.parseHTML would not make sure that the actual strings would be escaped; For example, if we get some result that's `http://foo.com/?"><script>alert('boo');</script>` that would cause an xss injection.

jQuery automatically escapes and does a couple of checks when you insert string values into the values. So, it would be ideal to escape *everything* but at the very least, link properties (like href) are more crucial, which is why I changed those right there.

src/RevisionPopupWidget.js
mooeypoo: Probably not? the inside (which we're not sure about) is parseHTML'ed, but the outside we wrote ourselves, so it's less crucial. The bigger issue is when we can't be 100% sure about the value of what we're getting -- which is true about the comment and the username.

src/RevisionPopupWidget.js
MusikAnimal: Firefox 52 according to MDN. But at any rate, I'm more interested if these should apply here, since they are internal links? I noticed the wiki does not use this for internal links.

Side note, all external links on the wiki have `noopener` and `noreferrer`, but I definitely still see referrer data in my tools' access logs saying they came from the wiki, and even for the latest browser versions. Maybe they're spoofing, who knows...

MusikAnimal: I'll certainly get this added to XTools, which has many an external link. Pageviews and Event Metrics are probably fine.

src/RevisionPopupWidget.js
MusikAnimal: The links are built with `mw.util` which does escaping anyway, no? Security is obviously important but I wonder if we're maybe going over board here. `<`, `>`, etc. aren't allowed in usernames to begin with, and the edit summary HTML comes from the API which also escapes markup.

src/RevisionPopupWidget.js
mooeypoo: Hm that's a good point, if they're strictly internal links, always, then you're right, it's probably an overkill, I didn't think about the fact that we know 100% that the links are always internal.

src/RevisionPopupWidget.js
mooeypoo: I don't think they're escaped with mw.util, but either way, we should *always* escape a template.

Also, why is this overboard? We have jQuery loaded anyways, and to be fair, this feels a little less cluttered than the templates especially after the later pending patches where we add quite a bunch to them, and they start getting really confusing. I think there's a place for templates, but it's okay to be a bit more vigilant when dealing with clickable things. Also, we can't be sure about what people put into usernames or comments even if it's not `>` and `<`, and escaping is always something that should to be done, regardless of anything, so we don't miss a use case. We skipped escaping the other content (I think we should, to be fair) but it was, at least, content we're a bit more sure about, but anything inside a clickable href should be a bit more higher scrutiny.

Honestly, I discovered this when I started adding the bidi encapsulation method into the templates and it started getting super confusing and cluttered.

The tempates were good when the content was clear and short, but we keep adding more and more to it, wrapping things around (bidi or security or whatever else) that some of those templates are become much less clear than a straight-up jquery object that, by merely existing, also solves a lot of potential worries about security, xss, and other issues jquery solves for us in the background.

If we already load it, and we're already using it, and it makes things a bit more organize to where we don't even need to consider whether it should or shouldn't be escaped, then why not?

MusikAnimal: #54 helps clean up the templating some. I personally find template literals much more readable, especially for simple text content; `document.createTextNode( ' ' + mw.msg( 'parentheses-start' ) )` for instance seems quite verbose. What we have here is 20 lines of logicless code that was just 6 before. I can't imagine what the code would look like if it were a whole page of jQuery-generated content!

Anyway I've no strong feelings :) The jQuery stuff isn't what I'm questioning, rather I'm confused what we're escaping that isn't already escaped. If we can't rely on MediaWiki to do this then I think we have much bigger problems.

```
> mw.util.getUrl('Special:Contributions/<script>')
"/wiki/Special:Contributions/%3Cscript%3E"
```
Is there another example of what we're worried about? If mw.util doesn't do proper escaping, I think it certainly should.

At any rate, the code is already written, so there's no reason to change it.

src/RevisionPopupWidget.js
MusikAnimal: Did we want to remove it, then? I don't think it's overkill, just curious if we wanted to match MediaWiki.

MusikAnimal (approved):
Works great! I am unable to find a mainspace edit summary in a RTL language, but I see that `<bdi>` is where it should be.

Just the one outstanding question about noopener/noreferrer, otherwise :+1: from me.

src/RevisionPopupWidget.js
mooeypoo: Yeah, I am not sure in this case. We're talking about internal links we control, so it's probably okay. If this wasn't a gadget or extension then the problem of google not indexing it would've actually been a problem, but that's not really the case here anyways.

So, yeah, I think this is a case where it's not actually crucial either way -- and we can probably remove it. I'll do that in the next patch fix.

BTW, To be fair, just specifically the "match" to MediaWiki, I think we (generally) might want to do it the other way around and start adding those to relevant (non internal) places in MW if/when we do target=_blank (which is already not common). Anyways, that's totally an aside. I'll raise it to Volker.

src/RevisionPopupWidget.js
mooeypoo: Yeah some of this came out of me trying to make sense of the templating in #54 while thinking about my new patch that needs to wrap things. Some of the new wrap functions made things more confusing rather than less, so I didn't think that changing only the links was much of a loss.

The `document.createTextNode( ' ' + mw.msg( 'parentheses-start' ) )` is only there because when I encapsulated the links, I thought it would be easier to encapsulate them all at once in a jQuery array, so Roan and I "shoved" the plaintext in there. We could, alternatively, create 3 jQuery links and combine them in a template -- which would get rid of the above. At that stage, though, it just didn't seem to matter *that* much. I could bring that back, though, if you think it looks a lot better?

In relation to the encapsulation -- I agree that in general terms, because we know exactly what we're getting from the DB and (hopefully, yeesh) from the `mw.util` methods, it's probably not insanely critical to escape things, but I feel like that is the "right way" in general when writing code for the web, so adding that security measure -- combined with the fact it's slightly more readable for the links -- felt better to me. 

I think that using templates is a matter of balance. They can make things much much much more clear, but they can also turn out to be a blob of "messy" string-in-var-in-string mess, and I think we just need to be careful of that extreme too. I started noticing this when I was looking at the other PRs and they started being a bit more complicated.

With #54 for example, I was wondering if (in the future?) we want to do something where the promise changes the DOM, which is clearer in concept (not necessary right now, but if we start adding more and more behavior, then that would make it consolidated into a single logical operation) -- and with the way that the template started to look super bloated, I was getting confused about the inline methods and the big string concatenations being full of `${xxx}` that was inside `${function(xxx)}` and near each other.

There's a general rule when building the DOM to try and avoid concatenating elements because the browser's DOM parser has its own (very good) way of preventing us from making mistakes. It's usually better to either use `document.createNode` or jQuery. That said, templates definitely have their place, but when things become dependent on a lot of sanitation/wrapping/methods I felt like it was time to re-assess.

Some of this is also personal preference over what each of us finds as "clearer" :D 
I think the best thing here is that you and I could continue to tug each other a bit, we can come up with a balance that we're both happy with -- which would likely also bring balance to unforseen issues with either approach.

mooeypoo:
> Works great! I am unable to find a mainspace edit summary in a RTL language, but I see that `<bdi>` is where it should be.

So, best examples of this would be using WWT on enwiki (or any of the available wikis) but set your interface language to an RTL language. You'll see that without the patch the punctuation (and a little bit of the alignment) is off without BIDI, and works much better with it. Also, I think there was a conversation about adding hewiki to whocolor (not sure?) so that would make it even more prevalent.

Anyways, I answered inline and I plan to fix-up the things we discussed a bit more, so I should have a new patch in ready for your re-review/merge tomorrow, hopefully. 

Thanks!

MusikAnimal (changes requested):
The selling point of template literals, for me, is seeing the raw markup with interpolations just as you would with a templating engine like Twig, Handlebars, etc., but thank you for trying to appease me :) The use of template literals as you have it here actually introduced extraneous spacing between the parentheses. You'd need to put the parentheses on the same line as the links inside of it.

I'm totally okay if you go back to what you had before. I think it was more readable than this version... Sorry for taking up so much of your time!

Here's an example of what I'm seeing. Note the space around the links:

Screenshot from 2019-09-30 20-17-16

Also, can you tell me if the `<bidi>` is working as it should in this example? I realize some Hebrew translations are missing, but it still looks quite off for RTL. This is what I see on the master branch:

Screenshot from 2019-09-30 20-22-02

mooeypoo:
Ha, we can't win! ;) 

Anyways, damn, I didn't notice the spacing. Argh. Okay, I've reverted back to the previous jQuery structure, but we can keep an eye on this while we work for future iterations, there might be an easier/nicer look to all of this that still takes into account a bit more of the wins we automatically get from jquery.

As for the behavior -- it looks good to me since we don't consider the parentheses part of the text but rather part of the interface language. So, the best way to test this would be with either a comment or a username that end with either numbers or punctuation.... I wonder if we can find some sample revision with that, I'll try to find something.

MusikAnimal (approved):
Merging! Thanks for putting up with me :) This was great code from the beginning, I again was mostly curious about what we were guarding against since we're working only with data given by MediaWiki and with MW utilities. But it all makes sense, better safe than sorry, and good to follow best practice if nothing else. :100:

PR #56, opened 2019-09-23
Fix highlighting cutting off font kerning

cimurah:
Add a ::before pseudo element to the 'active' CSS class, which is used to highlight when hovering
or selecting an element. When the .active class is set, a pseudo element is added behind it with
a background color that highlights the text. This behind-element approach will prevent any content
from being obscured.

Bug: T232490

### DIFF
--- a/src/less/general.less
+++ b/src/less/general.less
@@ -20,9 +20,20 @@ html.wwt-popup body {
 
 .mw-parser-output .editor-token {
 	cursor: pointer;
+	position: relative;
+	display: inline-block;
+	white-space: pre;
 
-	&.active {
+	&.active::before {
+		content: '';
 		background-color: #facc48;
+		position: absolute;
+		top: 0;
+		left: 0;
+		z-index: -1;
+		width: 100%;
+		height: 100%;
+		padding-bottom: 6px;
 	}
 }
 
--- a/src/less/general.less
+++ b/src/less/general.less
@@ -20,20 +20,10 @@ html.wwt-popup body {
 
 .mw-parser-output .editor-token {
 	cursor: pointer;
-	position: relative;
-	display: inline-block;
-	white-space: pre;
+	font-kerning: none;
 
-	&.active::before {
-		content: '';
+	&.active {
 		background-color: #facc48;
-		position: absolute;
-		top: 0;
-		left: 0;
-		z-index: -1;
-		width: 100%;
-		height: 100%;
-		padding-bottom: 6px;
 	}
 }
 


src/less/general.less
```diff
@@ -20,9 +20,20 @@ html.ext-wwt-popup body {
 
 .mw-parser-output .editor-token {
 	cursor: pointer;
+	position: relative;
+	display: inline-block;
```
mooeypoo: A little worried about setting all tokens as `inline-block` rather than `inline`, but if you don't see any negative impact, it's probably fine...? :D

cimurah: Below is what I've learned during my research :)

We need the inline element to behave as a block so that when an element starts with a space it renders in the same line.  Inline elements that start with a space and are at the end of a line, will start at the end of the line and then continue to the beginning of the next line.  The `::before` pseudo element with the background color will not follow the `inline` element, that they are attached to, to the next line. Here is a visual of the different types of display:  https://stackoverflow.com/questions/9189810/css-display-inline-vs-inline-block 
We are also able to add CSS attributes, like padding, for `inline-block` to cover the white lines between lines in paragraphs.

Do you foresee any possible issues with this solution?

samwilson: I think one of the issues with `inline-block` is that the whole thing will be wrapped to the next line sometimes, instead of broken in the middle. For example: https://codepen.io/freosam/pen/MWgxMPG This doesn't seem to be that much of an issue because most `editor-token` elements are next to others, but if they are next to e.g. an inline element introduced by a template then wrapping might go awry.

src/less/general.less
```diff
@@ -20,9 +20,20 @@ html.ext-wwt-popup body {
...
+	&.active::before {
+		content: '';
 		background-color: #facc48;
+		position: absolute;
+		top: 0;
+		left: 0;
+		z-index: -1;
+		width: 100%;
+		height: 100%;
+		padding-bottom: 6px;
```
samwilson: I'm getting a slightly odd display in Firefox, where the bottom padding goes downwards too far, e.g.:
highlight-overlap

This goes away with `top: -3px; padding-bottom: 3px` but I'm not sure what else that breaks.

samwilson (changes requested):
Looks great. I think my comment about inline templates is probably not worth worrying about too much, because the layout changes on load anyway (i.e. galleries, categories, etc. look different) so a bit of wrapping difference isn't likely to be too ugly.

src/less/general.less
mooeypoo: I'm good with iterating on this; if it works 99% of the time and the 1% edge case is super edgy and probably not going to be introduced, then I think that's fine. Especially since this fixes a bigger edge case, so it's an overall win.

We can start with this potential problem and if we see that it appears often, we can rethink this piece, it's fairly contained so it's not like we will have a lot of trouble iterating on it.

MusikAnimal (changes requested):
The `display: inline-block` looks pretty bad for image thumbnails :confused: Try for instance [street artist](https://en.wikipedia.org/wiki/Street_artist), or a more drastic example, [Banksy](https://en.wikipedia.org/wiki/Banksy). Then you have weird alignment issues; in every paragraph it looks like all but the first line has a bit of left-margin.

Screenshot from 2019-09-30 21-38-15

The requirement to have the highlighting touch adjacent lines seems problematic too. At least in Chrome on Ubuntu, it looks largely like spacing is only below each line and not equally above. This can overlap superscript on subsequent lines (e.g. search and hover over "Michael Bloomberg" on the Banksy article), and some of the highlighted area isn't actually clickable.

Screenshot from 2019-09-30 21-35-52

Overall I really question if this is the right approach. The kerning issue seemed relatively minor to begin with. Is fixing it worth all the repercussions?

No issue with the code itself, that all looks great :) I just wonder if visually we're taking a step backwards. Has @prtksxna looked this over?

mooeypoo:
It looks like the problem is double-sided. First, the `inline-block` creates some issues, but then also the superscript messes the line-height.

I created a quick experiment that doesn't use inline-block and pushes the superscript to not mess up the line-height, so the rendering is better.

https://codepen.io/mooeypoo/pen/KKPjZOe

This has some minor issues, but from my experiment, it might work better?

prtksxna:
As others have noted above, the highlight not being 'centered' on the  sentence looks kinda _off_. I am not sure what the trade-offs and technical difficulties are, but here are my 2¢:

1. The kerning issue (T232490) seems minor and doesn't affect usability much (unless someone is trying to figure out WWT for a period or comma and this issue is making it hard to hover on something). If it is a lot of work it is ok if this issue is not fixed.
2. It would be ideal if the highlight is centered. It being centered is more important than it being tall enough to fill gaps between sentences. So:

<table>
<tr>
<td>Nope </td>
<td>Ok</td>
<td>Great</td>
</tr>

<tr>
<td>
Screenshot 2019-10-05 at 6 07 22 PM
</td>
<td>
Screenshot 2019-10-05 at 6 09 01 PM
</td>
<td>
Screenshot 2019-10-05 at 6 08 39 PM
</td>
</tr>

</table>

***

Let me know if there was a different question that you actually wanted answered :P

MusikAnimal:
I think
```css
.mw-parser-output .editor-token {
    font-kerning: none;
}
```
will remove the kerning in non-Microsoft browsers (Edge will soon support it, too), if we're okay settling with that.

cimurah:
Added `font-kerning: none;` as suggested by @MusikAnimal to fix the kerning. I think is good to have this attribute added for the browsers that do support it. I removed the code that attempted to eliminate the white line between lines. We can address that at a later time if we find it necessary.

MusikAnimal (approved):
Looks good to me! I'm assuming others agree (since we did intend on removing the kerning), so I shall merge.

PR #57, opened 2019-10-01
Scroll down to content on activation

samwilson:
If the content is below or at the bottom of the viewport when
WWT is activated (e.g. on a diff page), scroll down to it. This
happens after the API request is received, because failures will
have no need to scroll.

Bug: https://phabricator.wikimedia.org/T232059

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -96,6 +96,7 @@ class App {
 					$contentWrapper.html( this.api.getReplacementHtml() );
 					$( 'body' ).append( this.revisionPopup.$element );
 					this.attachContentListeners();
+					this.scrollDown();
 					this.widget.setState( 'ready' );
 				},
 				// Error handler.
@@ -106,6 +107,23 @@ class App {
 			);
 		// Add a class for CSS to style links appropriately. Is removed in this.onWidgetClose().
 		activationInstance.getContentWrapper().addClass( 'ext-wwt-enabled' );
+	}
+
+	/**
+	 * Scroll down to the content on a diff page
+	 * if it is below or in the lower third of the viewport.
+	 */
+	scrollDown() {
+		const viewBottom = window.scrollY + ( window.innerHeight * ( 2 / 3 ) ),
+			$diffHead = $( 'h2.diff-currentversion-title' );
+		if ( $diffHead.length === 1 ) {
+			const contentTop = $diffHead.offset().top,
+				infobarHeight = this.widget.$element.outerHeight( true );
+			if ( contentTop > viewBottom ) {
+				// Scroll to below the WWT info bar. Redundant selector is for Safari.
+				$( 'html, body' ).animate( { scrollTop: contentTop - infobarHeight } );
+			}
+		}
 	}
 
 	/**


src/App.js
```diff
@@ -96,6 +97,17 @@ class App {
 					$contentWrapper.html( this.api.getReplacementHtml() );
 					$( 'body' ).append( this.revisionPopup.$element );
 					this.attachContentListeners();
+
+					// Scroll down to the content if it is below or in the lower third of the
+					// viewport, e.g. on a diff page.
+					viewBottom = window.scrollY + ( window.innerHeight * ( 2 / 3 ) );
+					contentTop = $contentWrapper.offset().top;
+					infobarHeight = this.widget.$element.outerHeight( true );
+					if ( contentTop >= viewBottom ) {
```
mooeypoo: Just an idea here, but from experience with rcfilters, scrolling is a bit awkward/disruptive, we might want to consider a buffer (or threshhold) where if you're already almost at the article we don't scroll, and we only scroll if it makes sense to. This will prevent cases where you can see the article halfway up the screen and we scroll you like 10 pixels down, which is weird, but will keep the behavior of scrolling when the article is not visible or obviously there.

Also, we might want to use jquery's `.animate()` so the scrolling is slightly less janky.

See example here with the `min: 20` https://github.com/wikimedia/mediawiki/blob/master/resources/src/mediawiki.rcfilters/ui/FilterTagMultiselectWidget.js#L415

The actual scrolling method: https://github.com/wikimedia/mediawiki/blob/master/resources/src/mediawiki.rcfilters/ui/FilterTagMultiselectWidget.js#L819

samwilson: I like the animation idea, that'd help people realise what's happening.

This is already only scrolling if the target is out of or almost out of view. The threshold is currently a third of the viewport height, but could be smaller (i.e. a shorter area at the bottom of the screen).

@prtksxna do you have an preference?

mooeypoo: I missed the 2/3 calculation, good thinking!

samwilson: I've added the animation.

mooeypoo:
This looks great to me; one thing I'd consider is that right now the scroll ends up with the title (`Revision as of xxxx`) being under the top bar, which is different than the way we load it in a non-diff page (the initial position of the bar is above the title)

Do we want to fix that (scroll to just above the title) or is this okay?

samwilson:
> Do we want to fix that (scroll to just above the title) or is this okay?

Good point!

We could stroll down to the top of the preceding sibling of the content (which is what the "Latest revision as of" `h2` is).

Although, are there any other situations like this other than the diff page? I can't think of any. So we could just scroll down to the h2 explicitly. Does that sound too limited?

mooeypoo:
> Although, are there any other situations like this other than the diff page? I can't think of any. So we could just scroll down to the h2 explicitly. Does that sound too limited?

I can't think of any other use case, but we can start with that --and if we find another use case, fix it up and adjust. I think that'll be the most straight forward way to go anyways?

samwilson:
I've updated it to scroll to the diff header if it's there, and otherwise to scroll to the content. Sound okay?

mooeypoo (approved):

PR #58, opened 2019-10-01
Support hidden usernames

MusikAnimal:
Show "(username or IP removed)" when the username has been hidden from
public view. We don't use the 'rev-deleted-user' message because, at
least on English Wikipedia, it uses parser functions and magic words
that aren't available in this context.

Highlighting tokens where the contributor username is hidden is still
not supported, but this can be addressed in a separate PR.

Bug: T232215


MusikAnimal:
Note this takes some code from #54. Doesn't matter which PR gets merged first.

mooeypoo (approved):

PR #59, opened 2019-10-02
Display a verbal message if percentage is less than 1%

mooeypoo:
In the revision popup, if an editor has a percentage that is less
than 1%, do not display the fraction, but instead display a literal
message saying 'less than 1%'.

Bonus for this PR: Add parameter documentation to qqq.json for the
original message.

Bug: https://phabricator.wikimedia.org/T233164


samwilson (approved):
Works for me!

MusikAnimal (approved):

PR #61, opened 2019-10-07
Always use correct revision ID

samwilson:
Get rid of URL parsing and instead use mw.config values. This not
only gets us away from parsing the URL manually but more
importantly means the revision ID will always be correct (we were
just using the 'oldid' URL parameter, which is incorrect when
there is also a 'direction' URL parameter).

https://phabricator.wikimedia.org/T232059

Bug: T232059

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -8,6 +8,7 @@ class Api {
 	* @param {Object} config
 	* @cfg config.url The WikiWho base URL.
 	* @cfg config.mwApi The mw.Api instance.
+	* @cfg config.mwConfig The mw.config data (a mw.Map object).
 	* @constructor
 	*/
 	constructor( config = {} ) {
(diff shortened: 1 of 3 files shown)


samwilson:
The only issue I can think of is that this uses `wgContentLanguage` instead of the subdomain, which is true for all of the wikis we're targeting but isn't universally true. We could instead extract it from `wgServerName`, but I dunno if it's worth it.

mooeypoo:
> The only issue I can think of is that this uses `wgContentLanguage` instead of the subdomain, which is true for all of the wikis we're targeting but isn't universally true. We could instead extract it from `wgServerName`, but I dunno if it's worth it.

I'd use what WhoColor is using; I think there might be differences between content language code and some wiki subdomain code. not 100% sure,but almost.

If that's the case, I'd just stick t othe URL, since that's also what WhoColor's using anyways.

samwilson:
> I'd use what WhoColor is using; I think there might be differences between content language code and some wiki subdomain code. not 100% sure,but almost.
> 
> If that's the case, I'd just stick t othe URL, since that's also what WhoColor's using anyways.

Done. Falls back to the lang code, but that's pretty unlikely to happen.

mooeypoo (approved):
Looks good!

PR #62, opened 2019-10-08
Remove 'ext-' prefix from messages and class names

samwilson:
https://phabricator.wikimedia.org/T234386


MaxSem:
In principe, you don't have to prepend `whowrotethat` either, just specify a prefix in the Translatewiki config.

samwilson:
That'd be even nicer! Does that still work even when we're injecting it into Wikipedia? Wouldn't there be clashes with other messages? Extensions and gadgets don't do it that way.

MaxSem:
Hmm, good question... Probably not :P

MusikAnimal (approved):

PR #63, opened 2019-10-08
MVC: Add a model and controller for system state

mooeypoo:
The primary purpose of this patch is to allow for WWT system to
launch and shut itself down through a unified operation. This is
so that we can make sure we shut the system down if another system
is at risk of interfering with it, and set the system to a disabled
state in a way where all the distinct parts are synchronized.

For this purpose, the following was done:
- Create a model that stores and manages the system's state, whether
  it is enabled, and emits events on change of any of those states.
- Rename activationSingleton to Controller (that's already what
  it's doing, so solidify that) and use it to manage all actions
- Make sure that widgets listen to model events and call the controller
  on actions.
- Aim for better separation of concerns; controller deals with direct
  actions. Model stores state. App handles widgets.
  The current implementation isn't perfectly separate, but it aims to
  be a lot more organized in this hierarchy and operation.
- Repurpose `App` to be the place that manages the content area behavior
  but leave the rest to the controller.
- Create a unified operation to launch/dismiss the system.
- Make sure that launch is dependent on whether the system is enabled.
  This will enable us to disable the system (for example, if VisualEditor
  is turned on) without too much juggling of external code.
- Add the launch/dismiss code to the wwtDebug routine so it can be
  triggered from the console.
- Redo relevant tests.

Bug: https://phabricator.wikimedia.org/T234874


src/Controller.js
```diff
@@ -107,11 +111,81 @@ class ActivationSingleton {
...
 
+	/**
+	 * Launch WWT application
+	 *
+	 * @return {jQuery.Promise} Promise that is resolved when the system
+	 *  has finished launching, or is rejected if the system has failed to launch
+	 */
+	launch() {
+		if ( !this.model.isEnabled() ) {
+			window.console.log( 'Who Wrote That: Could not launch. System is disabled.' );
```
MusikAnimal, line 129: I noticed some console.logs have `WhoWroteThat: ...` and others with spaces. Maybe we could refactor this out somehow, into Tools.js perhaps? The method would take just a message, and would prepend `WhoWroteThat` (or with spaces, doesn't matter).

src/InfoBarWidget.js
```diff
@@ -71,8 +76,9 @@ InfoBarWidget.static.legalFlags = [ 'pending', 'ready', 'err' ];
  * Change the state of the widget
  *
  * @param {string} state Widget state; 'pending', 'ready' or 'error'
+ * @param {string} errorCode Error code, if applicable
```
MusikAnimal: Should `errorCode` be `[errorCode]`? I'm not actually sure.

src/Model.js
```diff
@@ -0,0 +1,184 @@
...
+			 *
+			 * @event Model#enabled
+			 * @type {boolean}
+			 * @param {boolean} isEnabled System is currently active
+			 */
+			this.emit( 'enabled', this.enabled );
+		}
+	}
+
+	setState( state, errorCode = '' ) {
```
MusikAnimal, line 207: Does this need a JSDoc? Following https://github.com/wikimedia/WhoWroteThat/commit/420acd2e5501885bd771b3fc55e347c9d27f2bb7#diff-1dc6ee56b778cd91e0327b52aaeaa1b9R21 the tests should fail. Also `isInitialized` above, and `getApi` and `getModel` in Controller.js, though it's quite obvious what those methods do :smiley:

src/InfoBarWidget.js
```diff
@@ -32,12 +32,17 @@ const InfoBarWidget = function InfoBarWidget( config = {} ) {
...
+	this.setState( wwtController.getModel().getState() );
 	this.toggle( false );
-	this.setLabel( $( '<span>' ).append( mw.msg( 'ext-whowrotethat-state-pending' ) ).contents() );
 
-	// Close event
-	this.closeIcon.$element.on( 'click', () => this.emit( 'close' ) );
+	// Events
+	this.closeIcon.$element.on( 'click', () => {
+		wwtController.dismiss();
+	} );
```
MusikAnimal: This perhaps could be simplified to `this.closeIcon.$element.on( 'click', wwtController.dismiss );`

src/Model.js
```diff
@@ -0,0 +1,184 @@
...
+
+	/**
+	 * Cache the original content we are about to replace
+	 *
+	 * @param {jQuery} [$content] A jQuery object to cache. If not given
+	 *  the $contentWrapper will be cached. In most cases this
+	 *  should remain empty, as $contentWrapper contains the
+	 *  original content before it is replaced.
+	 */
+	cacheOriginal( $content ) {
```
MusikAnimal: Does `cacheOriginal( $content = this.$contentWrapper )` work?

src/Model.js
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	/**
+	 * Check if the current page is valid
+	 * for the system to display.
+	 *
+	 * @return {boolean} Page is valid
+	 */
+	isValidPage() {
+		return !!(
```
MusikAnimal, line 95: The `&&`s below should cast this to a Boolean anyway, no?

src/Controller.js
```diff
@@ -20,33 +23,38 @@ class ActivationSingleton {
 			this.lang = 'en';
 			this.translations = {};
 
-			ActivationSingleton.instance = this;
+			Controller.instance = this;
```
MusikAnimal, line 23: I can't comment on lines that aren't part of the diff; but it looks like `this.$contentWrapper` and `this.$originalContent` are no longer being used (lines 21-22)?

src/Controller.js
```diff
@@ -107,11 +111,81 @@ class ActivationSingleton {
...
+	/**
+	 * Toggle the system; if already launched, dismiss it, and vise versa.
+	 */
+	toggle() {
+		if ( this.model.isActive() ) {
+			this.dismiss();
+		} else {
+			this.launch();
+		}
+	}
```
MusikAnimal, line 208: New line below this? There is an eslint rule for this I'm sure, if want to use it.

MusikAnimal (commented):
I left some general feedback (non-blocking). `wwtDebug.launch()` and `wwtDebug.dismiss()` appear to work flawlessly :) I take it there are no other user-facing changes.

src/Controller.js
mooeypoo: That's an awesome idea. Should I do it in this one, or a followup? I don't want to do too many to make this too big, but ... maybe if I'm already moving things around it's a good time? Not sure.

src/InfoBarWidget.js
mooeypoo: Whoops, yes!

src/Model.js
mooeypoo: Oops. Yes. Hah. Not sure how I forgot those.

src/InfoBarWidget.js
mooeypoo: Good point -- though should be `wwtController.dismiss.bind( wwtController );`

src/Controller.js
MusikAnimal: Follow-up is totally fine. In a recent script I was working on, I also allowed the method to accept a log level (`log` (default), `warn`, and `error`). Don't know if we'd want to do the same.

src/Model.js
MusikAnimal: I'd like to know why the eslint rule didn't work... I will investigate!

src/Model.js
mooeypoo: Yeah I was thinking about that, wasn't sure it'll work... but then I tested this in the console and it seemed to work!

```
class Foo {
    constructor() {
       this.bar = 'test';
    }

    test( x = this.bar ) {
       return x;
    }
};
const f = new Foo();
console.log( 'f.test()', f.test() );
console.log( 'f.test("something else")', f.test( 'something else' ) );
f.bar = 'changed test';
console.log( 'f.test()', f.test() );

> f.test() test
> f.test("something else") something else
> f.test() changed test
```

src/Controller.js
mooeypoo: Good point, I moved these to the model.

Also, I think that we can get rid of the singleton signature... it's not really needed since we control the entire code and we're not actually creating a new instance, so there's no real need to fortify against ourselves.

I was trying to be careful how much I rewrite though, so we can do that as a followup.

src/Controller.js
mooeypoo: Yeah let's add that rule, I'll check what it is. I thought it was already in there, but maybe not for es6 modules rules...?

src/Controller.js
```diff
@@ -0,0 +1,257 @@
...
+	 * Initialize if it is the first time, and cache the
+	 * initialization for the next times.
+	 */
+	constructor() {
+		if ( !Controller.instance ) {
+			this.initialized = false;
+			this.link = null;
+			this.namespace = null;
+			this.mainPage = false;
+			this.lang = 'en';
```
samwilson: Both the model and the controller have a lang attribute; could the controller just use the model's directly? i.e. add a `getLang()` method to the model? Seems that there are two places currently defining the default of 'en' (which is immediately overwritten anyway, but still).

Can totally be done in a follow-up though.

.jsdoc.json
```diff
@@ -35,6 +35,7 @@
         "jQuery.Promise": "https://api.jquery.com/Types/#Promise",
         "Object": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object",
         "Promise": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise",
+		"EventEmitter": "https://nodejs.org/api/events.html",
```
samwilson: Indent with spaces, not tabs.

src/Model.js
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	/**
+	 * Toggle the active state of the system
+	 *
+	 * @fires Model#active
+	 * @param {boolean} [isActive] System is active
+	 */
+	toggleActive( isActive ) {
+		isActive = isActive !== undefined ? !!isActive : !this.active;
```
samwilson: Maybe this could have a comment... it's a bit of a mouthful.

src/Model.js
```diff
@@ -0,0 +1,215 @@
...
+		return this.state;
+	}
+
+	/**
+	 * Toggle the active state of the system
+	 *
+	 * @fires Model#active
+	 * @param {boolean} [isActive] System is active
+	 */
+	toggleActive( isActive ) {
```
samwilson: Instead of `isActive` maybe `newState`, to make it clearer that it's the *target* state, not the current state?

src/Model.js
```diff
@@ -0,0 +1,215 @@
...
+		}
+	}
+
+	/**
+	 * Toggle whether the system is currently enabled
+	 *
+	 * @fires Model#enabled
+	 * @param {boolean} [isEnabled] System is enabled
+	 */
+	toggleEnabled( isEnabled ) {
```
samwilson: Same as for toggleActive.

src/Controller.js
```diff
@@ -0,0 +1,257 @@
...
+		}
+
+		return Controller.instance;
+	}
+
+	getModel() {
+		return this.model;
+	}
+
+	getApi() {
```
samwilson, line 59: Needs a docblock.

src/Controller.js
```diff
@@ -0,0 +1,257 @@
...
+			// Events
+			this.model.on( 'enabled', isEnabled => {
+				this.getButton().toggle( isEnabled );
+			} );
+		}
+
+		return Controller.instance;
+	}
+
+	getModel() {
```
samwilson, line 50: Needs a docblock.

samwilson (changes requested):
Looks great! This is going to be much clearer. It works great in my testing. I just have a couple of pretty minor comments.

In reviewing, I also think I found a but with the jsdoc markdown plugin. :)

src/Controller.js
mooeypoo: Yes! Good catch. The model should be the one storing this.

.jsdoc.json
mooeypoo: Ugh

src/Controller.js
mooeypoo: egh we really need to add that to eslint

samwilson (approved):

PR #65, opened 2019-10-09
Add a consistent logging tool

mooeypoo:


src/Tools.js
```diff
@@ -2,6 +2,33 @@
  * Class to hold some global helper tools
  */
 class Tools {
+	/**
+	 * Output into the console.
+	 *
```
samwilson, line 9: Maybe worth commenting here that there's another `log()` function in `build/extension_content_script.js`, so future us know to update in both places if required.

build/extension_content_script.js
```diff
@@ -1,4 +1,14 @@
 var theBrowser = chrome || browser;
+
+/**
+ * Unify the way we output console.info
+ * We have this tool in Tool.log but that code is not yet loaded when this file runs.
```
samwilson: Should be `Tools.log()`, plural.

src/Tools.js
```diff
@@ -2,6 +2,33 @@
...
+	 * @param  {string} [type='info'] Log type used for console output
+	 *  valid options are:
+	 *  - 'info' for console.info()
+	 *  - 'warn' for console.warn()
+	 *  - 'error' for console.error()
+	 * @param  {string} str String to output
+	 * @param {Mixed} [params] Parameters to output into the console
+	 *  alongside the message. This is especially useful for errors.
+	 */
+	static log( type = 'info', str = '', params ) {
```
samwilson: Not a big thing, but as there are five instances of calling this with 'info' and one of 'error', maybe make the type parameter come 2nd, e.g. `log( str, type = 'info', params = null )` and then not pass 'info' for the five.

src/Tools.js
```diff
@@ -2,6 +2,33 @@
...
+		const prefix = 'Who Wrote That extension: ';
+		let method = window.console.info;
+		if ( type === 'warn' ) {
+			method = window.console.warn;
+		} else if ( type === 'error' ) {
+			method = window.console.error;
+		}
+
+		if ( params ) {
```
samwilson: It doesn't apply with the current usage of this argument (the API result), but if someone wanted to output a false or null value it wouldn't be caught here. Maybe check for undefined instead.

src/Tools.js
```diff
@@ -2,6 +2,33 @@
...
+	 *  valid options are:
+	 *  - 'info' for console.info()
+	 *  - 'warn' for console.warn()
+	 *  - 'error' for console.error()
+	 * @param  {string} str String to output
+	 * @param {Mixed} [params] Parameters to output into the console
+	 *  alongside the message. This is especially useful for errors.
+	 */
+	static log( type = 'info', str = '', params ) {
+		const prefix = 'Who Wrote That extension: ';
```
samwilson: Question mark missing?

build/extension_content_script.js
mooeypoo: Whoops

src/Tools.js
mooeypoo: Yeah good point, I'll change that

src/Tools.js
mooeypoo: Oh yeaaah....

src/Tools.js
```diff
@@ -2,6 +2,33 @@
...
 class Tools {
+	/**
+	 * Output into the console.
+	 *
+	 * @param  {string} [type='info'] Log type used for console output
+	 *  valid options are:
+	 *  - 'info' for console.info()
+	 *  - 'warn' for console.warn()
+	 *  - 'error' for console.error()
+	 * @param  {string} str String to output
```
MusikAnimal: `[str]`? Though I think requiring a value for `str` might be better.

src/Tools.js
MusikAnimal: I might suggest `debug` as the default level. Most all of this seems to be for that purpose, and this would avoid log spam.

src/Tools.js
```diff
@@ -2,6 +2,33 @@
...
+			method = window.console.warn;
+		} else if ( type === 'error' ) {
+			method = window.console.error;
+		}
+
+		if ( params ) {
+			method( prefix + str, params );
+		} else {
+			method( prefix + str );
+		}
```
MusikAnimal, line 30: You might could simplify lines 19-30 with `window.console[type]( prefix + str, params )`, setting an empty string as the default value for `params` on line 17. The idea here is that `log( 'foo', 'whatever' )` is the same as `console.whatever( 'foo' )`, whereby we trust we use something valid for the type.

src/Tools.js
mooeypoo: Yeah and we should probably examine whether the `error` we're outputting on the API error is actually an error (it just notifies us, it's more debug) especially since I noticed Chrome considers it an error in the extension log :(

src/Tools.js
mooeypoo: Haaa good point

src/Tools.js
mooeypoo: Ooh! Good point! I can also just validate the types just-in-case but that's not a must.

samwilson (approved):
Looks good to me!

PR #66, opened 2019-10-10
Remove jQuery from bundled script

mooeypoo:
We don't need to bundle jQuery module into the script because
we are injecting it into MW where jQuery already exists.

Remove jQuery from babelify operation, and make sure that the module
only loads if `$` isn't set, which is relevant for tests.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -97,21 +97,24 @@ module.exports = function Gruntfile( grunt ) {
 				src: 'src/outputs/gadget.js',
 				dest: 'dist/gadget/generated.whowrotethat.js',
 				options: {
-					transform: [ 'babelify' ]
+					transform: [ 'babelify' ],
+					exclude: [ 'jquery' ]
 				}
 			},
 			browserextension: {
 				src: 'src/outputs/browserextension.js',
 				dest: 'dist/extension/js/generated.pageScript.js',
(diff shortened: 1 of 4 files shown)


samwilson:
Hooray!

This means we can turn on warningsAsErrors in package.json too.

samwilson (approved):

samwilson:
I did wonder about setting `window.$` earlier and then just always using it in the different classes, rather than duplicating the check for the global. But nah. This works. :-)

mooeypoo:
The main thing about the window.$ is to name sure we only load the jQuery module for tests, otherwise they fail 😔

samwilson:
But could we set `window.$` in the test before requiring Controller etc.?

mooeypoo:
So, I tried that. I removed the `const $ = window.$ || require( 'jquery' );` from Tools.js and Controller.js and instead put `window.$ = require( 'jquery' );` at the top of the Tools.test.js and Controller.test.js files and I get this error when running `grunt test`

```
ReferenceError: $ is not defined
    at Suite.$ (/home/moriel/code/workcode/WhoWroteThat/test/suite/Controller.test.js:6:23)
    [...]
```
:disappointed:

samwilson:
Ah well! How it is is good. :)

PR #68, opened 2019-10-10
Require JSDoc blocks for all methods

MusikAnimal:


mooeypoo:
Unfortunately, this means that the pr should also include the fixes (adding missing doc blocks) or it won't pass the test 😂

We can do a preliminary one where we add the missing blocks first and then enable this maybe?

MusikAnimal:
> Unfortunately, this means that the pr should also include the fixes (adding missing doc blocks) or it won't pass the test 

Yes of course! I submitted it to advertise what still needs documenting. I didn't write any of these methods, so hoping I could get some help so that I don't say the wrong thing.

I'm also still trying to figure out why it doesn't fail for prototype function expressions, e.g. `InfoBarWidget.prototype.clearUsernameInfo`.

mooeypoo:
Great, added all the missing ones in #69

PR #69, opened 2019-10-10
Add missing jsdoc block comments

mooeypoo:


MusikAnimal (approved):
You could have copied #68 here (or pushed to that branch), but this works too! Thank you :)

PR #71, opened 2019-10-12
Cache all promises and DOM elements

mooeypoo:
This cleanup makes sure we are always working with the current and correct data, and are intentionally adding events once to the DOM we are getting from the API, without having to redo the operation in case WWT is closed and reopened.

However, this also makes sure that if the revision has changed between close/reopen, we are then refetching the correct data and caching it.

Specifically in this commit:
* Cache the WhoColor API data based on the revisionID we are currently looking at; do not refetch unnecessarily, refetch when needed.
* Cache revision summary API requests based on revision ID. For followup, this will also allow us to control the animationn better and decide to animate only if the data has not yet been fetched.
* Always send revision ID to WhoColor API. We can make 100% sure that what we get from WhoColor is the specific revision we're looking at. This protects us from two main cases:
  1. Case where we look at the recent revision, but while we are looking at the article and before we asked WWT for info, someone else edited the same article. In the case before this commit, WhoColor would have sent us the latest revision which would have been different than what we are looking at.
  2. If a user edited the page in VisualEditor or the new wikitext editor, the page does not refresh. If we ask for WWT data, we want to make sure we're getting the correct new revision ID rather than an older one.
* Create a sanitized token info object on initialization. This means that we have the data available immediately to render on hover and click, rather than having to constantly look it up in the API results. Also, this allows us to make the API process more internally independent, where we only cache data that is needed rather than having a lot of the methods having to go back-and-forth with class variables that are set within the Promise.
* Have `App` class fully handle the DOM we receive from the API. By storing it properly, we make sure we always deal with a fully processed DOM without having to attach events over and over, unless the DOM element has explicitly changed because the revision ID we are looking at has changed. This also allows us to be much safer in what DOM we are returning back to the user on deactivation, and make sure we encapsulate the behavior so we can switch the elements back and forth safely.

This commit will make it possible, as followups, to:
* Control the 'loading' animation of the edit summaries; since they are now based on cached promises, it will be easier to only display the loading animation only if needed.
* On VisualEditor save, it is now possible to refetch WWT data, whether it exists or not in WhoColor.

Bug: https://phabricator.wikimedia.org/T234483

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -19,7 +19,10 @@ class Api {
 
 		this.mwApi = config.mwApi;
 		this.mwConfig = config.mwConfig;
-		this.results = null;
+		this.promiseCache = { data: {}, summaries: {} };
+		this.tokenMap = {};
+		this.maxRetries = 4;
+		this.retry = 1;
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


src/Api.js
```diff
@@ -86,37 +107,70 @@ class Api {
...
 		// account the 'direction' URL parameter.
 		revId = this.mwConfig.get( 'wgRevisionId' );
-		curRevId = this.mwConfig.get( 'wgCurRevisionId' );
-		if ( revId !== curRevId ) {
-			parts.push( revId );
-		}
+		// Always include the revision ID, to make sure we are always asking for
+		// the correct revision, whether the page was just edited, or, whether the
+		// page was edited by someone else while we were looking at the current page
+		parts.push( revId );
```
samwilson: This addition to the `parts` array can be moved up to where the others are; it was only down here because it was conditional.

src/Api.js
```diff
@@ -86,37 +107,70 @@ class Api {
...
+	 * Get a unique identifier for a specific promise.
+	 * The promise is unique if the revision ID is the same, and should
+	 * change (and hence force a refetch) if the revision ID changed, like
+	 * in the case of looking at the page after a VE save (without refresh)
+	 *
+	 * @return {string} Identifier for this promise
 	 */
-	getReplacementHtml() {
-		return this.results.extended_html;
+	getContentRevisionIdentifier() {
```
samwilson: I was going to suggest that this method be used above in `getAjaxURL()`, but maybe it's meant to be a *cache* identifier, not a revision identifier (I mean, they might be the same thing of course). If it is then maybe rename this method, and if it's not then it can be used in both places and the `mwConfig.get( 'wgRevisionId' )` only occur here.

samwilson (commented):
With this patch there seems to be something odd happening: when I launch, immediately close, and then re-launch WWT I get a hover colour and then nothing (no further hovering or clicking functions at all). No error messages.

I'm still looking into what's going wrong. Is anyone else seeing this?

src/Api.js
mooeypoo: ooh, good point

mooeypoo:
> With this patch there seems to be something odd happening: when I launch, immediately close, and then re-launch WWT I get a hover colour and then nothing (no further hovering or clicking functions at all). No error messages.
> 
> I'm still looking into what's going wrong. Is anyone else seeing this?

Hmmm I didn't see this when I was testing, but this sounds like I'm not replacing the original/current DOM properly. I think it might need to be replaced properly *after* everything is done, perhaps in a `.done` or something. I'll take another look.

src/Api.js
mooeypoo: Aye, it's a cache identifier inside the API, but I'm also using it to check whether things changed in the controller... It is basically the identifier for the current version we have cached, so if the content changed, we compare to see whether it's the same as the cache-identifier, and if it's different, we ask for a refetch.

 Should I rename it?

samwilson: I think either call it `getRevisionId` and change the docblock accordingly, or call it `getCacheKey` and leave it as-is. It's totally minor though! I just got confused reading the code above and wondered why this class is calling `this.mwConfig.get( 'wgCurRevisionId' )` twice.

test/suite/Model.test.js
```diff
@@ -118,7 +118,14 @@ describe( 'Model test', () => {
...
 		} );
 		it( 'getOriginalContent returns original content', () => {
-			expect( m.getOriginalContent().html() )
+			expect(
+				// We want to examine the *actual* html of the
+				// original content; but .html() gives us everything
+				// except the top node, so we wrap with a mock
+				// $('<div>').html() so that the inner representation
+				// is what we actually sent into the original content
+				$( '<div>' ).append( m.getOriginalContent() ).html()
```
samwilson: This could also be done as `m.getOriginalContent()[ 0 ].outerHTML`.

samwilson (commented):
I'm still trying to figure out why it's hanging for me when I enable it twice (it's still doing it, but not on every page).

src/Controller.js
```diff
@@ -166,7 +174,7 @@ class Controller {
 			return this.api.getData( window.location.href )
 				.then(
 					// Success handler.
-					() => {
+					result => {
```
samwilson, line 177: This patch fails for me when you launch, close-before-load, then launch again before the first load has finished (in the background). This is because, now the same promise is being returned from `Api.getData()`, there are two `.then()` invocations being branched on the same promise (i.e. not chained). The order in which they execute isn't determined, but the first one will do what we want, and the second one will break things. This is why, if you launch, close, then relaunch after 10 seconds or something, everything will work ok.

This means in practice that the 2nd invocation returns the tokenized HTML, but without the listeners, and so hovering is broken.

A (possibly naive) fix is to check at this point what state we're in, and not return the new HTML if we're already `ready`:

```javascript
if ( this.model.getState() === 'ready' ) {
	return;
}
```

This works in my testing.

samwilson (changes requested):

src/Controller.js
mooeypoo: Ha! GOOD CATCH!!

I've added it to the condition that also checks if the model is active. Good one!

Also, this isn't entirely just a naive fix; the operation, at its end, sets the state to "ready" so it makes sense that it only runs if the state *isn't* already ready. Good catch.

test/suite/Model.test.js
mooeypoo: Done.

src/Api.js
mooeypoo: Done.

src/Api.js
mooeypoo: Done.

mooeypoo:
I fixed the bug (thanks @samwilson !! ) and the other comments.
Please re-review :)

samwilson (approved):

PR #72, opened 2019-10-15
Change string-built HTML to jQuery objects

samwilson:
Bug: T232802

### DIFF
--- a/src/RevisionPopupWidget.js
+++ b/src/RevisionPopupWidget.js
@@ -25,11 +25,13 @@ OO.inheritClass( RevisionPopupWidget, OO.ui.PopupWidget );
 /**
  * Get markup for the diff size.
  * @param {number} size
- * @return {string}
+ * @return {jQuery}
  */
 function getSizeHtml( size ) {
-	let sizeClass;
+	const $diffBytes = $( '<span>' );
+	$diffBytes.addClass( 'mw-diff-bytes' );
 
+	let sizeClass;
 	if ( size > 0 ) {
 		sizeClass = 'mw-plusminus-pos';
 	} else if ( size < 0 ) {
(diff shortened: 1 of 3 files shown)


src/RevisionPopupWidget.js
```diff
@@ -125,14 +128,12 @@ RevisionPopupWidget.prototype.show = function ( data, $target ) {
...
-			${attributionMsg}
-		` );
-
-	this.$popupContent.html( html );
+		$attributionMsg = $( '<div>' )
+			.addClass( 'wwt-revisionPopupWidget-attribution' )
+			.html( mw.message( scoreMsgKey, data.score ).parse() ),
+		$html = $( '<div>' )
+			.append( addedMsg, getCommentHtml( data ), $attributionMsg );
+	this.$popupContent.html( $html.html() );
```
mooeypoo: At this point, you can just do `this.$popupContent.empty().append( ... )`

src/RevisionPopupWidget.js
```diff
@@ -25,48 +25,51 @@ OO.inheritClass( RevisionPopupWidget, OO.ui.PopupWidget );
...
-				<div class="wwt-shimmer www-shimmer-animation"></div>
-			</div>`;
+		const $shimmerDiv = $( '<div>' );
+		$shimmerDiv.addClass( 'wwt-shimmer www-shimmer-animation' );
+		$revCommentDiv.append( $shimmerDiv, $shimmerDiv );
+	} else {
+		$revCommentDiv.addClass( 'wwt-revisionPopupWidget-comment-transparent' );
+		const $commentSpan = $( '<span>' );
+		$commentSpan.addClass( 'comment comment--without-parentheses wwt-revisionPopupWidget-comment' );
+		$commentSpan.append( Tools.bidiIsolate( data.comment ) );
```
mooeypoo: you can chain those ...

```
const $commentSpan = $( '<span>' );
    .addClass( ... )
    .append( ... )
$revCommentDiv
    .addClass( ... )
    .append( ... );

Comment (mooeypoo) [inline src/RevisionPopupWidget.js]: Also, `Tools.bidiIsolate` needs to be re-fortified. Right now, it automatically html's any comment we get, which isn't great.

It should go back to only outputting literal strings as .text() which forces the callers to demonstrate the safety of sent data, like the API comment.

File: src/RevisionPopupWidget.js [hunk 3]
```diff
@@ -25,48 +25,51 @@ OO.inheritClass( RevisionPopupWidget, OO.ui.PopupWidget );
...
 	} else {
 		sizeClass = 'mw-plusminus-null';
 	}
+	$diffBytes.addClass( sizeClass );
 
-	return `
-		<span class="${sizeClass} mw-diff-bytes">` +
-			`${size > 0 ? '+' : ''}${mw.language.convertNumber( size )}` +
-		'</span>';
+	$diffBytes.text( `${size > 0 ? '+' : ''}${mw.language.convertNumber( size )}` );
```
mooeypoo: You can combine those two lines

```
return $diffBytes
    .addClass( sizeClass )
    .text( ... );
```

src/RevisionPopupWidget.js
MusikAnimal: Sorry for the mess I created! I just wanted to note that the edit summary can contain internal links and we'll want to render that HTML. Whether that works with `.text()` I'm not sure.

src/RevisionPopupWidget.js
samwilson: Done.

src/RevisionPopupWidget.js
samwilson: Done.

src/RevisionPopupWidget.js
samwilson: Done.

src/RevisionPopupWidget.js
samwilson: @mooeypoo it seems like that just reduces Tools.bidiIsolate to:

```javascript
	static bidiIsolate( $content ) {
		return $( '<bdi>' ).append( $content );
	}
```

i.e. if we make it only accept a string or jQuery object, and only return jQuery object.

We can't get away from having to parse the HTML of the comment, but we can move it out of `bidiIsolate()`.

@MusikAnimal I don't think it's a mess! But yeah, I guess we're really aiming to never create HTML with strings any more.

mooeypoo: Okay, two things here, following up [from the ticket](https://phabricator.wikimedia.org/T232802#5558253) --

1. Demonstrable security: By making the tool *itself* force people to explicitly send it jQuery objects when they intend to do so, it makes sure that those choices are not only intentional, but also that they *look* intentional to anyone reading the code.
2. With the current behavior, we cannot send a raw text that we want to *not* parse as html, ever, which is bad. An example would be a username; we actually never want to html-parse a username, and the fact that it might be safe(ish) 90% of the time doesn't mean we shouldn't fortify against it. 

When you look at the code, this:

```
Tools.bidiIsolate( $( '<span>' ).html( data.comment ) );
```
Tells you that we intentionally html-parse comments, and this --

```
Tools.bidiIsolate( data.username );
```
Tells the code reader that they don't need to worry what username has inside it, whether it's justified or not, because it will be displayed as `.text()` which is always safe.

We shouldn't underestimate perceived security, and we should fortify ourselves against html-parsing things we really shouldn't.

(BTW, while it's true that usernames *should* be safe, it's also worth noting that the array of illegal characters was somewhat slowly added to Wikipedia, and we never actually removed past usernames that used them, which means we might actually have situations where usernames *are not* html-safe.)

samwilson: Of course, good points. I've switched it to using `.text()` for strings, and appending all else.

samwilson:
We might want to do away with `.html()` for messages, but I'll do that in a separate patch.

PR #75, opened 2019-10-16
Separate Firefox and Chrome manifests

samwilson:
This adds the Gecko ID back into manifest.json, and creates a
new build process that uses separate manifest files for Chrome
and Firefox, outputting two zip files for upload to the web
stores.

We removed the Gecko ID because I thought it wasn't required and
it was causing a warning in Chrome. Turns out the storage.sync API
isn't available unless you set it.

https://phabricator.wikimedia.org/T236226
https://phabricator.wikimedia.org/T232460

Bug: T236226
Bug: T232460

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -164,20 +164,35 @@ module.exports = function Gruntfile( grunt ) {
 		clean: [ 'dist/', 'temp/' ],
 		shell: {
 			mocha: 'mocha --require @babel/register test/testHelper.js --recursive test/suite  --colors',
-			webextBuild: 'web-ext build',
+			// Build two extension zip files for uploading to the browser stores.
+			// Note that the manifest.json in the Chrome zip file is actually from
+			// manifest_chrome.json in the dist directory.
+			webextBuild: [
+				// Build for Firefox.
+				'web-ext build --artifacts-dir dist/extension_firefox',
+				// Swap manifests, and build for Chrome. The manifest_chrome.json has already been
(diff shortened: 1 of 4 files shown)


samwilson:
This patch will introduce a warning in Chrome, as it stands. I'm looking at generating two extension directories in `dist/`, one for Chrome and one for Firefox. Chrome only allows installation of local extensions from a directory with a manifest file, otherwise I'd say we could generate two zip files instead.

Gruntfile.js
```diff
@@ -162,7 +162,17 @@ module.exports = function Gruntfile( grunt ) {
 		clean: [ 'dist/', 'temp/' ],
 		shell: {
 			mocha: 'mocha --require @babel/register test/testHelper.js --recursive test/suite  --colors',
-			webextBuild: 'web-ext build',
+			webextBuild: [
```
mooeypoo, line 170: OK I'm going to be a little petty, but this is really confusing and I want to make sure we know what this is doing even when we get back to this in 6 months :stuck_out_tongue_winking_eye: 

Let's document here that 
- The chromeManifest job _runs first_
- This specific job ends up with files in `dist/` and also 2 zip files for the extensions
- Optionally, emphasize that there's a difference between the manifest in `dev` and the zipped one. It doesn't matter in practice, but we should remember that when/if debugging something.

samwilson: Okay, I've added some more comments. I've also moved the source zip file into the `extension_firefox` directory, so it's not hanging out on its own in the directory above.

PR #76, opened 2019-10-20
Move HTML out of i18n messages

samwilson:
Create a new Tools.i18nHtml() function for inserting jQuery HTML
elements into MediaWiki i18n messages.

https://phabricator.wikimedia.org/T232802

Bug: T232802


i18n/en.json
```diff
@@ -5,19 +5,23 @@
...
 	"whowrotethat-ready-general": "Hover to see contributions by the same author. Click for more details.",
-	"whowrotethat-tour-welcome-title": "<strong><em>Who Wrote That?</em></strong> is enabled!",
+	"whowrotethat-tour-welcome-title": "$1 is enabled!",
+	"whowrotethat-tour-welcome-title-name": "Who Wrote That?",
 	"whowrotethat-tour-welcome-description": "Activate the tool from the sidebar. Hover over the text to see the author and their other contributions to this page. Click to get more details.",
 	"whowrotethat-tour-welcome-dismiss": "Got it!",
 	"whowrotethat-revision-added": "$1 added this on $2.",
-	"whowrotethat-revision-attribution": "They have written <strong>$1%</strong> of the page.",
-	"whowrotethat-revision-attribution-lessthan": "They have written <strong>less than 1%</strong> of the page.",
+	"whowrotethat-revision-attribution": "They have written $1 of the page.",
```
mooeypoo, line 24: I guess this isn't really the fault of this PR since it was there before, but "They" should be {{GENDER}} and we should send the username in for the gender. Maybe as a followup.

src/Tools.js
```diff
@@ -60,5 +60,33 @@ class Tools {
...
+			marker = '@@I18N',
+			// Create a list of parameter markers.
+			replacements = Object.keys( params ).map( function ( k ) { return marker + k; } );
+		// Insert the markers into the message, in place of $1, $2, etc.
+		let outputMsg = mw.msg.apply( null, [ msg ].concat( replacements ) );
+		// Replace each marker in the message with the actual HTML.
+		replacements.forEach( function ( rep, idx ) {
+			const param = params[ idx ],
+				// Parameters can be string or jQuery.
+				val = typeof param === 'string' ? param : param[ 0 ].outerHTML;
```
mooeypoo: Are we not risking the same problem we had before, where raw `<script>...</script>` in translated messages are actually inserted as HTML even if they're strings? They only difference between those two is that one would insert the raw string (that could be html) and one would insert the string representing outer html; both would then be parsed as HTML when injected into the output.

Am I missing something?

mooeypoo: I think the way to "fix" this is to do the same trick we do in `bidiIsolate` -- instead of outputting a string in this method that is then `.append( ... )` into jquery (which could mean we're injecting HTML we aren't controlling) -- we have this method always output either a jQuery object (parsed) or a jQuery.text() (which is always safe)

That would mean that whenever we do `.append( Tools.i18nHtml(...) )` we can always trust that we're appending safe html.

test/suite/Tools.test.js
```diff
@@ -39,4 +39,53 @@ describe( 'Tools test', () => {
 			} );
 		} );
 	} );
+
+	describe( 'Tools.bidiIsolate', () => {
+		const cases = [
+			{
```
mooeypoo, line 41: Just to make a point to our reviewers (and ourselves 😄 ) I'd add a test here that shows that any raw string with html in it is being outputted safely; something like `<em>foo</em> => &lt;em&gt;foo&lt;/em&gt;`

src/Tools.js
```diff
@@ -60,5 +60,33 @@ class Tools {
...
+	 * @param {jQuery|string} params jQuery objects or strings to insert into the message.
+	 * @return {string}
+	 */
+	static i18nHtml( msg, ...params ) {
+		const
+			// Temporary marker string that will be replaced (with an integer appended).
+			// Must not appear in messages.
+			marker = '@@I18N',
+			// Create a list of parameter markers.
+			replacements = Object.keys( params ).map( function ( k ) { return marker + k; } );
```
mooeypoo: I wasn't going to make this a blocker, but since I have another comment below, I'm going to channel @MusikAnimal and recommend you ES6'ify this 

`.map( k => marker + k );` should work, I think.

samwilson: Good point. Done.

src/Tools.js
samwilson: You're quite right. I've replaced the string that's injected with `$( '<div>' ).text( param ).html()`, which escapes it. We could of course use `mw.html.escape()` but that's not available in tests; do you think it's worth mocking it? It seems easier to use jquery but I don't really mind.

The thing about string parameters here is that we don't ever actually use them at the moment. I only added support for them because it seems like it'd be nice to introduce this sort of thing into jquery.i18n (oh yes, that's the other reason to avoid mw.html).

test/suite/Tools.test.js
samwilson: Done.

i18n/en.json
samwilson: Good catch. Will make another PR.

samwilson (commented):
Thanks! I feel a bit silly for missing the obvious. :-)

PR #77, opened 2019-10-24
Add locale files for the browser extension

samwilson:
This means the name and description shown in the browser web
stores will be localized.

Bug: T236226

### DIFF
--- a/build/extension_manifest.json
+++ b/build/extension_manifest.json
@@ -1,8 +1,9 @@
 {
 	"manifest_version": 2,
-	"name": "WhoWroteThat for Wikipedia",
+	"name": "__MSG_name__",
+	"description": "__MSG_description",
+	"default_locale": "en",
 	"author": "Wikimedia Foundation",
-	"description": "@@description",
 	"homepage_url": "https://meta.wikimedia.org/wiki/Community_Tech/Who_Wrote_That_tool",
 	"version": "@@version",
 	"icons": {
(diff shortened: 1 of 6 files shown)


build/extension_manifest.json
```diff
@@ -1,8 +1,9 @@
 {
 	"manifest_version": 2,
-	"name": "WhoWroteThat for Wikipedia",
+	"name": "__MSG_name__",
+	"description": "__MSG_description",
```
mooeypoo, line 4: Sorrry, I'm confused here. Where are these __MSG_xx things being replaced? or... set up? 

I was used to the `@@xxx ` pattern of replacements, and this looks odd. Is this a part of the extension definition standard?

samwilson: Yep! This is confusing, that's why you're confused. It's good like that. :)

So, the `@@xxx` replacements are made when this file is copied to `dist/extension/`, but these `__MSG_xxx__` ones are *not*. They stay like this, even in the zip file that's uploaded to the stores, and then the stores do some magic to produce what's actually sent to the browsers downloading the extensions.

This makes it possible to view the extension page on addons.mozilla.com in different languages.

samwilson: See also https://phabricator.wikimedia.org/T236226#5597488

mooeypoo: Gotcha, okay, so this is basically the message-key "prefix" for the store. Is this true to both mozilla and chrome?

It would be more important when we do the Chrome-specific deploy but I am just curious. Anyways, it's non-blocking. This looks good!

PR #80, opened 2019-10-29
Combine WelcomeTour into same context, and check for load

mooeypoo:
We need to have the WelcomeTour be able to check whether or not the
model is enabled and whether VE is being loaded on the page.

When the user navigates to a page that is already in ve-edit mode,
the way that the load order goes prevents the different parts of
the system to trust that they can immediately (or rationally) tell
whether VE is loading, so the popup was injected and displayed as
the model was updated.

Further complicating things is the fact that babel seems to write
a separate instantiation of the Controller and Model when the
entry files are separated merely by the fact it is creating local
variables that are each within its own `<script>` tag. This meant
that when the main loading sequence updated the Model's enabled
state, the WelcomeTour could not see it because it received a
different context instantiation of the model.

To fix those issues, this commit:

- Moves the WelcomeTour to be inside the instantiation of the
  main system, with conditionals set on whether to run it.
  This makes sure the tour and system use the same model.
- The model's initial 'enabled' state is set to false, and is
  only enabled if ve is not actually in the process of loading.
- The tour popup is only displayed if VE is not being loaded, and
  awaits the state of the model to be displayed.
  Product wise, this means that the popup will not display and then
  vanish when VE loads, and also that when VE is closed (with or
  without refresh) the popup the shows up, once per session, until
  dismissed.

Bug: https://phabricator.wikimedia.org/T231508


src/outputs/browserextension.js
```diff
@@ -15,6 +15,71 @@ config.outputEnvironment = 'Browser extension';
...
+			$.when(
+				$.ready,
+				mw.loader.using( [
+					'oojs-ui',
+					'mediawiki.jqueryMsg'
+				] )
+			).then( function () {
+				// The check about whether the tour was already seen has already
+				// taken place at this point. This whole file would not be injected
+				// by the contentScript to begin with, if WelcomeTourSeen is true
```
samwilson: "This whole file" is no longer true, is it?

samwilson (commented):
Works in my testing. Looks good. I think I follow the motivations. :-)

I'm having some issues testing in Firefox because of the storage permission issue. Could we merge #75 before this one?

PR #82, opened 2019-10-30
Build different zip files for beta and prod

samwilson:
The name and description are different for prod and beta, so this
adds new i18n messages for these and dynamically adds them to the
web extensions' locales files. It also separates extension
building from the main build task (into a new 'webext' task).

Bug: T236226

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -138,6 +138,8 @@ module.exports = function Gruntfile( grunt ) {
 			manifest: {
 				options: {
 					patterns: [
+						// The trailing '0' here is also used in the compress.webextSource Grunt
+						// task below.
 						{ match: 'version', replacement: pkg.version + '.0' }
 					]
 				},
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -138,6 +138,8 @@ module.exports = function Gruntfile( grunt ) {
 			manifest: {
 				options: {
 					patterns: [
+						// The trailing slash here is also used in the compress.webextSource Grunt
```
MusikAnimal: You mean the trailing `.0`?

MusikAnimal (changes requested):
Works great! Just one tiny issue.

Gruntfile.js
samwilson: The phrase "trailing slash" is too deep in my fingers' muscle memory, obviously!

Fixed. :-)

PR #83, opened 2019-10-31
Gray out non-tokenized elements

MusikAnimal:
Non-tokenized elements have no data associated with them, so we gray
them out to give visual indication of this. Hovering over the grayed-out
elements will increase the opacity to 80%.

There doesn't appear to be a rock-solid solution to identifying any and
all untokenized elements. So we loop through the DOM and pick out ones
that don't have .editor-token as a child. Additionally we have a
manually maintained list of selectors under `untokenizedElements` in
config.js, which covers some elements that are nested deeper than the
first level. This is not comprehensive but can be updated as needed.

Bug: T235130

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -1,7 +1,14 @@
 const config = {
 	wikiWhoUrl: 'https://www.wikiwho.net/',
 	// Set the output type between the extension and the gadget
-	outputEnvironment: ''
+	outputEnvironment: '',
+	// Elements that we know for sure are never tokenized and should be grayed out.
+	untokenizedElements: [
+		'.citation',
+		'.IPA',
+		'.mw-editsection',
+		'table'
+	]
 };
 
 // Set up a prefix for console outputs
--- a/src/App.js
+++ b/src/App.js
@@ -1,3 +1,4 @@
+import config from './config';
 import InfoBarWidget from './InfoBarWidget';
 import RevisionPopupWidget from './RevisionPopupWidget';
 import wwtController from './Controller';
@@ -63,6 +64,7 @@ class App {
 		// only what's actually inside it
 		this.$content = $( $.parseHTML( html ) ).contents();
 		this.attachContentListeners( this.$content );
+		this.grayOutUntokenizedElements( this.$content );
 	}
 
 	/**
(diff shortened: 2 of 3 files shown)


MusikAnimal:
Note some things don't have any wrapper around them, or a unique CSS class, so we can't dim them. For instance the ISBN template under "References" at https://en.wikipedia.org/wiki/Leekes, or the Arabic translation in the lede of https://de.wikipedia.org/wiki/Proteste_im_Libanon_2019.

I did also try graying out everything and un-graying out `.editor-token`. Due to how things can be nested I couldn't get this to work.

src/config.js
```diff
@@ -1,7 +1,16 @@
...
-	outputEnvironment: ''
+	outputEnvironment: '',
+	// Elements that we know for sure are never tokenized and should be grayed out.
+	untokenizedElements: [
+		'.citation',
+		'.IPA',
+		'.mw-cite-backlink',
+		'.mw-editsection',
+		'.reference',
+		'table'
```
samwilson, line 10: I'm not really sure why, but on Firefox, a table caption (like 'Club statistics' on [this page](https://en.wikipedia.org/wiki/Kevin_Ram%C3%ADrez_(footballer,_born_1993))) is not greyed out. It works correctly when I add 'caption' to this list, but the opacity goes weird (because it's compounding, I think).

src/config.js
```diff
@@ -1,7 +1,16 @@
...
 	// Set the output type between the extension and the gadget
-	outputEnvironment: ''
+	outputEnvironment: '',
+	// Elements that we know for sure are never tokenized and should be grayed out.
+	untokenizedElements: [
+		'.citation',
+		'.IPA',
+		'.mw-cite-backlink',
+		'.mw-editsection',
+		'.reference',
```
samwilson: This greys-out footnote superscript links, which do work. Is this by design?

MusikAnimal: Yeah I wasn't sure... I grayed them out because they aren't tokenized (as in the `<ref>` tags themselves), but as you say the links work so no need to pretend they don't.

src/config.js
MusikAnimal: I'm not sure what to do here. I tried `caption, tr` and `table > *` etc., but you either end up with the compounding opacity or the `:hover` acts on just the `tr`s / `caption`s, when (I think) we'd rather the hover work on the whole table as a single entity, if that makes sense.

I found this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1286803 There are arguments on whether `caption` should get the same opacity as the parent, but the bug is still open after 3 years so I take it that it is in fact a bug. Maybe we could live with it? There are surely other elements that don't get grayed out either (in any browser), though I guess captions are fairly common.

In working on this I did find situations where there was compounded opacity (e.g. table inside of a table), and I think I have that fixed.

PR #90, opened 2019-11-09
Add different extension IDs for Firefox

samwilson:
Add the prod and beta extension IDs in manifest.json when
building for Firefox.

Bug: T237757

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -231,10 +231,23 @@ module.exports = function Gruntfile( grunt ) {
 		} );
 	} );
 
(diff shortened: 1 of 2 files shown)


MusikAnimal (approved):
I think this works... I see the extension.json for Firefox prod uses whowrotethat@wikimedia, and the Beta version uses the ID. I don't fully understand the system but this appears to do what T237757 describes, so I'll merge :)

With `grunt webext` I do get warnings about "message must be 132 characters or less" but as I understand this is a known issue unrelated to this PR.

PR #93, opened 2019-11-14
Build: Add locale output to `grunt build`

mooeypoo:


samwilson (approved):
Makes the full build a tiny bit longer (because it duplicates the messages processing for prod) but we don't mind that. :)

PR #94, opened 2019-11-21
Fix percentage attribution score for IP contributors

cimurah:
Fix user edit scores for IP contributors shown in details
pop-up. The score should match the data retrieved from WhoColor Api.

Bug: T235775

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -178,10 +178,6 @@ class Api {
 		revision = results.revisions[ revId ];
 		username = revision[ 3 ];
 
-		// WikiWho prefixes IP addresses with '0|'.
-		isIP = username.slice( 0, 2 ) === '0|';
-		username = isIP ? username.slice( 2 ) : username;
-
 		// Get the user's edit score (percentage of content edited).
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1563,8 +1563,7 @@
          "version": "1.2.0",
          "resolved": "https://registry.npmjs.org/aproba/-/aproba-1.2.0.tgz",
          "integrity": "sha512-Y9J6ZjXtoYh8RnXVCMOU/ttDmk1aBjunq9vO0ta5x85WDQiQfUF9sIPBITdbiiIVcBo03Hi3jMxigBtsddlXRw==",
-         "dev": true,
-         "optional": true
+         "dev": true
```
samwilson: Is this change due to a different npm version?

But anyway, as you're not changing `package.json`, you can just remove `package-lock.json` from this commit.

cimurah: Not sure, but removed the file. Thank you!

samwilson (approved):

PR #101, opened 2019-11-28
Show size diff even when there's no revision comment

samwilson:
In the revision popup, show the size change (inline) even when
no revision comment has been given.

Bug: T236702


MusikAnimal (approved):
Works great.

The diff size fades in only if it has a summary to go with it. This can seem a tiny bit weird, because at quick glance some popups having content that fades in, and others don't. Maybe it makes sense to fade in `.mw-diff-bytes` in addition/separate to the full summary? That might look a little weird, and also dirty up the code. I think what you have here is more than fine. I just wanted to point out this tiny detail. Merging!

PR #103, opened 2019-12-03
Prevent content script from being injected more than once

samwilson:
Although I haven't quite figured out what's causing it, there are
times when the content script is injected twice. This adds a
guard against this, and even if it's not tackling the root cause
it shoudln't be too harmful a check to add.

This also removes the 'content_scripts.all_frames' setting from
manifest.json, returning it to its default value of false (meaning
that the script will only be loaded for the topmost frame in a
tab). This shouldn't change anything, because Wikipedia never
loads itself in a frame, but it seemed like a good guess as to
why things were loading twice (they still were even without it...).

https://phabricator.wikimedia.org/T232460

Bug: T232460

### DIFF
--- a/build/extension_content_script.js
+++ b/build/extension_content_script.js
@@ -10,18 +10,25 @@ function log( str ) {
 	window.console.info( 'Who Wrote That? (Browser extension): ' + str );
 }
 /**
- * injectScript - Inject internal script to available access to the `window`
+ * Inject the extension's content script to the DOM, so that it's got access to the `window` object.
  *
  * @param  {string} filePath Local path of the internal script.
  * @param  {string} tag The tag as string, where the script will be append (default: 'body').
  * @see    {@link http://stackoverflow.com/questions/20499994/access-window-variable-from-content-script}
  */
 function injectScript( filePath, tag ) {
-	var node = document.getElementsByTagName( tag )[ 0 ],
-		script = document.createElement( 'script' );
-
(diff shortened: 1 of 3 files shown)


mooeypoo:
I didn't know you can have `id=` field on a `<script>` tag! ha. Also, good catch.

build/extension_content_script.js
```diff
@@ -10,18 +10,25 @@ function log( str ) {
...
+	const wwtScriptId = 'wwt-content-script',
+		existingScript = document.getElementById( wwtScriptId );
+	if ( existingScript ) {
+		// Already injected.
+		return;
+	}
+	// Otherwise, inject a new script element.
+	let script = document.createElement( 'script' ),
+		node = document.getElementsByTagName( tag )[ 0 ];
+	script.id = wwtScriptId;
```
mooeypoo, line 29: `script` is an element; why not use `.setAttribute( ... )` like it did before? isn't the direct access to the properties less safe (in terms of behavior)?

samwilson: As I mentioned in Slack, I had a memory that best practice was to avoid `setAttribute()` in most cases, but I forgot why I remembered that. [This SO answer](https://stackoverflow.com/questions/10280250/getattribute-versus-element-object-properties) has some explanation and points to [more info](https://javascript.info/dom-attributes-and-properties).

PR #105, opened 2019-12-03
Don't gray out tables as they can contain tokenized elements

MusikAnimal:
Bug: T235130


mooeypoo:
Looks good, but please add a note about that in the CHANGELOG for the unrelease version so we don't forget to include that when the release comes out :)

mooeypoo (changes requested):
(Written in the comment)

Please add this change to the Changelog; let's try to start adding those as we do/merge them to save time *and* attach the change to the changelog line for easier git-blaming in the future :)

PR #106, opened 2019-12-04
Re-initialize the model on VE deactivation

samwilson:
VE replaces the entire content element, rather than just its
contents, and so our reference to it is no longer part of
the DOM. Re-initializing gives the model a reference to the
new element.

https://phabricator.wikimedia.org/T235015

Bug: T235015

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Prevent the extension script from loading more than once on a Wikipedia page
 - Don't gray out tables as they can contain tokenized elements
 - Use our standard debug log format (for VE activation/deactivation messages)
+- Correctly re-initialize WWT after VE deactivation
 
 ### Changed
 - Translation updates
--- a/src/Controller.js
+++ b/src/Controller.js
@@ -145,6 +145,11 @@ class Controller {
 			} );
 			mw.hook( 've.deactivationComplete' ).add( () => {
 				Tools.log( 'VisualEditor deactivated, enabling system.' );
+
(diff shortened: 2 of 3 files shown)


mooeypoo (approved):
Nicely done! Looks good, but haven't tested yet.

Should we add bug fix to the change log or is this not really a bug fix since the feature wasn't actually quite out yet?

I'll leave it to you to decide...

samwilson:
> Should we add bug fix to the change log or is this not really a bug fix since the feature wasn't actually quite out yet?

Done.

samwilson:
Is there any way that this re-initialization will stuff something up, though? I've tested, and can't see anything. But can you have a closer look too? (Tomorrow or whatever!) :)

mooeypoo:
I actually wondered that too; the *intended* usage was that it wouldn't break anything, but I was actually concerned that `model.initialize()` may be only partially updating what we need and that the initialization of the Controller does a bit too, but I just went over the code again -- and no, it should be safe.

The controller initialization does the actual init stuff like attaching button and classes and hooks, which should be once-per-session anyways. The model only stores stuff and emits events, so re-initializing it is fine. The concern I had was about re-caching the original content (because we basically replaced the original) but that's done every time the Controller launches: https://github.com/wikimedia/WhoWroteThat/blob/4a9e0d323a870df488d48fb9dfdeb41d88d1230d/src/Controller.js#L222

The only big deal here is the `initialized` event; the model will emit it every time it's re-initialized. Right now, that event is not used anywhere, but I'd maybe tweak the event documentation to make it clear that it *can* be done more than once, and that it means resetting the dom+config.

samwilson:
I've added a comment to `Model.initialize()`. Should be go to go now.

PR #109, opened 2019-12-05
Move click handler from button to the link inside the button

samwilson:
This might seem like an overly-large change for this little tweak,
but most of what it's doing is making the nomenclature for the
button and the link more consistent. The button is the `<li>` and
the link is the `<a>` inside it.

This patch also moves the creation of the portlet into the
getButton(), so we don't have the odd create-then-retrieve flow
in initialize(). And it turns both button and link into jQuery
objects, so we don't have to think too hard about whether they're
jQuery or HTMLElements (could have gone the other way, and done
them both as HTMLElements, but we want to use jQuery's toggle()
on them, so this is less code).

https://phabricator.wikimedia.org/T239842

Bug: T239842

### DIFF
--- a/src/Controller.js
+++ b/src/Controller.js
@@ -16,7 +16,8 @@ class Controller {
 	constructor() {
 		if ( !Controller.instance ) {
 			this.initialized = false;
-			this.link = null;
+			this.$button = null;
+			this.$link = null;
 			this.namespace = null;
 			this.mainPage = false;
 			this.translations = {};
@@ -125,14 +126,6 @@ class Controller {
 				)
 			);
 
-			// Add a portlet link to 'tools'
-			this.link = mw.util.addPortletLink(
-				'p-tb',
-				'#',
-				mw.msg( 'whowrotethat-activation-link' ),
-				't-whowrotethat',
-				mw.msg( 'whowrotethat-activation-link-tooltip' )
-			);
 			// Initial state
 			this.getButton().toggle( this.model.isEnabled() );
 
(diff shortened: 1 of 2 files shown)


src/Controller.js
```diff
@@ -311,12 +303,37 @@ class Controller {
...
+	 * Will add it if it doesn't already exist in the DOM.
 	 *
-	 * @return {jQuery} Activation button
+	 * @return {jQuery} Activation button.
 	 */
 	getButton() {
-		return $( this.link );
+		const portletId = 't-whowrotethat';
+		// If it's already been added to the DOM, return it.
+		if ( $( '#' + portletId ).length === 1 && this.$button instanceof $ ) {
```
mooeypoo: Why would this.$button ever be anything other than null (and then we create it) and jQuery instance? Do we need to check the existence of the portlet? Isn't it enough to just do `if ( this.$button ) return this.$button;` ?

src/Controller.js
```diff
@@ -311,12 +303,37 @@ class Controller {
...
+		return this.$button;
+	}
+
+	/**
+	 * Get the activation link, from inside the activation button.
+	 *
+	 * @return {jQuery} Activation link.
+	 */
+	getLink() {
+		return this.getButton().find( 'a' );
```
mooeypoo: If we're using this a lot, let's cache it internally. `.find( .. )` isn't entirely expensive, but it is a little bit; we could do something like

```
this.$button = $( mw.util.addPortletLink(...) );
this.$link = this.$button.find( 'a' );
```

and then `getLink()` can `return this.$link`  just so we don't have to redo the .find() again and again

mooeypoo (changes requested):

src/Controller.js
samwilson: Good point. Fixed.

src/Controller.js
samwilson: Done.

samwilson (commented):
Thanks, good ideas.

MusikAnimal (approved):
Works great!

I think something like "wrapper" or "container" might be a better name than "button", but this can be addressed in another PR (or ignored, doesn't matter :).

samwilson:
> I think something like "wrapper" or "container" might be a better name than "button", but this can be addressed in another PR (or ignored, doesn't matter :).

Yes, I actually changed it to `portletItem` and then back again, because I didn't want the diff to be too big. I might go back and do it though.

PR #111, opened 2019-12-10
Don't gray out elements with no visible content

MusikAnimal:
This is primarily to exclude empty `<p>` elements (possibly containing
only a `<br/>`), which is common in MediaWiki.

Bug: T235130


MusikAnimal:
This could very well have unintended side effects.

mooeypoo (changes requested):
This looks good overall, but I think you should cache `$( el )` in a variable and use it on the calls so jQuery doesn't reprocess it each time.

MusikAnimal:
> This looks good overall, but I think you should cache `$( el )` in a variable and use it on the calls so jQuery doesn't reprocess it each time.

Done!

samwilson (approved):

PR #112, opened 2019-12-10
Fix 2nd shimmer in revision popup

samwilson:
The second shimmer line was not being added correctly because
$.append() saw it as the same element. Cloning it fixes this.

Bug: T236702


MusikAnimal (approved):
Works! Though I realized we're no longer fading in the summaries. That doesn't appear to be the fault of this PR, though, and I'm not sure how important it is.

MusikAnimal:
> Though I realized we're no longer fading in the summaries. That doesn't appear to be the fault of this PR, though, and I'm not sure how important it is.

Raised at https://phabricator.wikimedia.org/T236702#5730148

PR #116, opened 2019-12-12
Fix revision details popup for image thumbnails

MusikAnimal:
Look for .editor-token elements that are immediate children of the
parser output, and not just further descendants.

Bug: T239112


mooeypoo:
Good catch!

PR #117, opened 2019-12-12
Add loading loop to wait for mw.loader.using to be available

samwilson:
This is because the Resource Loader queue is first processed by
the startup module, and it seems that we (sometimes only) get in
too quickly and it's not ready. This is a bit of a hack.


samwilson:
No longer required, as the cause of the error here has been fixed in MediaWiki.

PR #118, opened 2019-12-12
UX improvements for Infobar

MusikAnimal:
Bug: T234710

### DIFF
--- a/src/InfoBarWidget.js
+++ b/src/InfoBarWidget.js
@@ -19,7 +19,6 @@ const InfoBarWidget = function InfoBarWidget( config = {} ) {
 		flags: [ 'invert' ],
 		classes: [ 'wwt-infoBarWidget-close' ]
 	} );
-	this.userInfoUsernameLabel = new OO.ui.LabelWidget();
 	this.userInfoLabel = new OO.ui.LabelWidget( {
 		label: mw.msg( 'whowrotethat-ready-general' ),
 		classes: [ 'wwt-infoBarWidget-info' ]
@@ -49,7 +48,6 @@ const InfoBarWidget = function InfoBarWidget( config = {} ) {
 			this.$pendingAnimation,
 			this.$icon,
 			this.$label,
-			this.userInfoUsernameLabel.$element,
 			this.userInfoLabel.$element,
 			this.closeIcon.$element
 		);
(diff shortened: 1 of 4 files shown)


MusikAnimal:
@prtksxna Would you mind looking over this too? The only thing I had problem with was the _Move username closer to 'Who Wrote That?'_ part. It's using flexbox so I sort of eye-balled it. I do think it looks better:

Before:
Screenshot from 2019-12-12 17-35-22
Screenshot from 2019-12-12 17-35-33

After:
Screenshot from 2019-12-12 17-20-13
Screenshot from 2019-12-12 17-20-37

I tried playing with the `flex-grow` properties but couldn't come up with ratios that shifted the username further to the left, or reliably kept it at a certain placement. Would we prefer to use more finite widths, and do away with flexbox? Dare I say `table-cell`? :nauseated_face:

mooeypoo:
There are ways to set strict sizing for flex box too, and combine it with margins. I would stay away from table cell styling in css though. That is, indeed, 🤢

samwilson:
I think it's because not all siblings are flex items.

mooeypoo:
Now that we've hidden the username, we only have 2 active sections in that bar; the "Who Wrote That" logo/name and the help info with the [x]

Is this PR still relevant as a fix given the removal of the username?

src/InfoBarWidget.js
```diff
@@ -19,7 +19,10 @@ const InfoBarWidget = function InfoBarWidget( config = {} ) {
 		flags: [ 'invert' ],
 		classes: [ 'wwt-infoBarWidget-close' ]
 	} );
-	this.userInfoUsernameLabel = new OO.ui.LabelWidget();
+	this.$label.addClass( 'wwt-infoBarWidget-label' );
+	this.userInfoUsernameLabel = new OO.ui.LabelWidget( {
```
mooeypoo: This might not be relevant again now that the username is not in the bar

src/less/InfoBarWidget.less
```diff
@@ -126,8 +126,16 @@
 	&.oo-ui-flaggedElement-ready {
 		background-color: #36c;
 
-		.wwt-infoBarWidget-info.oo-ui-labelWidget {
+		.wwt-infoBarWidget-label {
+			flex-grow: 1;
+		}
+
+		.wwt-infoBarWidget-username {
```
mooeypoo: Same here; no longer used, no?

mooeypoo (changes requested):
Wrote 2 places that use the username (that isn't showing anymore) to get rid of / fix. Other than that, looks good. Might also solve the UX questions :)

prtksxna:
> @prtksxna Would you mind looking over this too? The only thing I had problem with was the _Move username closer to 'Who Wrote That?'_ part. It's using flexbox so I sort of eye-balled it. I do think it looks better:

Thanks for tagging me and sorry for such a late reply 😓

> After:
> Screenshot from 2019-12-12 17-20-13
> Screenshot from 2019-12-12 17-20-37

This definitely looks better. What I was imagining was that the username (or IP) should always start at the same position.


> I tried playing with the `flex-grow` properties but couldn't come up with ratios that shifted the username further to the left, or reliably kept it at a certain placement. Would we prefer to use more finite widths, and do away with flexbox? Dare I say `table-cell`? 🤢

Yeah finite width, especially for the "Who Wrote That?" text. I don't know much about the implementation here.

PR #121, opened 2019-12-13
Show revision details popup on top of infobar

MusikAnimal:
Bug: T234709

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -25,7 +25,10 @@ class App {
 
 		this.model = wwtController.getModel();
 
-		this.revisionPopup = new RevisionPopupWidget();
+		const $overlay = wwtController.getOverlay();
+		this.revisionPopup = new RevisionPopupWidget( {
+			$overlay
+		} );
 		this.widget = new InfoBarWidget();
 
 		// Attach widget
@@ -36,7 +39,8 @@ class App {
 		}
 
 		// Attach popup
-		$( 'body' ).append( this.revisionPopup.$element );
+		$( 'html' ).addClass( 'wwt-popup' );
+		$overlay.append( this.revisionPopup.$element );
 
 		// Attach events
 		this.$content = null;
--- a/src/Controller.js
+++ b/src/Controller.js
@@ -18,6 +18,7 @@ class Controller {
 			this.initialized = false;
 			this.$button = null;
 			this.$link = null;
+			this.$overlay = null;
 			this.namespace = null;
 			this.mainPage = false;
 			this.translations = {};
@@ -341,6 +342,21 @@ class Controller {
 		}
 		this.$link = this.getButton().find( 'a' );
 		return this.$link;
+	}
+
+	/**
+	 * Get the overlay div with the class wwt-overlay
+	 * Will add it if it doesn't already exist in the DOM.
+	 *
+	 * @return {jQuery} Overlay div
+	 */
+	getOverlay() {
+		if ( this.$overlay instanceof $ ) {
+			return this.$overlay;
+		}
+		this.$overlay = $( '<div>' ).addClass( 'wwt-overlay' );
+		$( 'body' ).append( this.$overlay );
+		return this.$overlay;
 	}
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Show revision details popup on top of infobar.
 
 ## [0.12.0] - 2019-12-19
 ### Fixed
--- a/src/RevisionPopupWidget.js
+++ b/src/RevisionPopupWidget.js
@@ -3,20 +3,24 @@ import Tools from './Tools';
 /**
  * @class
  * @constructor
+ * @param {Object} [config={}] Configuration options
  */
-const RevisionPopupWidget = function RevisionPopupWidget() {
+const RevisionPopupWidget = function RevisionPopupWidget( config = {} ) {
 	this.$popupContent = $( '<div>' )
 		.addClass( 'wwt-revisionPopupWidget-content' );
 	// Parent constructor
-	RevisionPopupWidget.parent.call( this, {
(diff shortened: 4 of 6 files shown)


mooeypoo (changes requested):
At this point, let's create an overlay for the popup to live in. This will make sure the popup is always on top of everything else in all skins and prevent us from running after random css rules that gadgets may use.

This is a common pattern in mediawiki Popups, I can find some examples of that'll help.

mooeypoo:
Just as a resource, here's how we do it normally within MediaWiki:

1. Create a new `<div>`, give it a class (norms dictate something like `wwt-overlay` or something like that)
2. Append this div to the end of the `<body>` tag so it's "on top" of everything else.
3. Append the popup $element to the overlay div you created.
4. Add this style to the overlay:
```
	position: absolute;
	right: 0;
	left: 0;
	z-index: 1;
```
See these examples:
- Echo overlay https://github.com/wikimedia/mediawiki-extensions-Echo/blob/cd6a340ce8b8899db9ce3ca11d2cd4f330a9ff41/modules/mobile/overlay.less#L14 
- OOUI Window overlay https://github.com/wikimedia/oojs-ui/blob/b752140988188b46ce890742417b605f4d449ce5/src/styles/Window.less#L34 

This will ensure the popup is always on top of everything, including gadgets or things we run around trying to fix. It's what we use for most popups in production, including Echo (as you see in the example)

MusikAnimal:
As discussed on Slack, this can happen but we're going to explore that as a separate issue:

Screenshot from 2019-12-19 14-18-19

src/App.js
```diff
@@ -36,7 +39,9 @@ class App {
 		}
 
 		// Attach popup
-		$( 'body' ).append( this.revisionPopup.$element );
+		$( 'html' ).addClass( 'wwt-popup' );
+		$( 'body' ).append( $overlay );
+		$overlay.append( this.revisionPopup.$element );
```
mooeypoo, line 43: As we talked about before, I was sure this should've happened within the widget itself since you're sending it your `$overlay`, but alas -- it turns out, ooui only appends the widget to the given overlay (in popups) in the `PopupButtonWidget` (https://github.com/wikimedia/oojs-ui/blob/a4f8b009c7a44aeed0d00dd43ead08e7ffdc0e79/src/widgets/PopupButtonWidget.js#L56)

I might check into whether the popup should do it to itself in ooui, but that's an aside.

src/App.js
```diff
@@ -25,7 +25,10 @@ class App {
 
 		this.model = wwtController.getModel();
 
-		this.revisionPopup = new RevisionPopupWidget();
+		const $overlay = $( '<div>' ).addClass( 'wwt-overlay' );
```
mooeypoo: If the welcome tour is active, we're already creating an overlay and attaching it to the body (see `browserextension.js` line 36). 

I'm wondering if it won't be easier to consolidate those two; it can even be in the form of creating the overlay as we set things up, and then pass the same overlay div to both of those; that way if the welcome popup appears we don't have two overlays attached to the DOM.

mooeypoo (changes requested):
Looks good... except one small thing. :)

src/App.js
cimurah: I consolidated the two. An overlay was added to `Controller.js`so that only one overlay div is passed around and attached to the DOM.

src/App.js
```diff
@@ -25,7 +25,10 @@ class App {
 
 		this.model = wwtController.getModel();
 
-		this.revisionPopup = new RevisionPopupWidget();
+		const $overlay = wwtController.getOverlay();
+		this.revisionPopup = new RevisionPopupWidget( {
+			$overlay
```
samwilson, line 30: The RevisionPopupWidget doesn't actually use $overlay (as @mooeypoo says below), so there's no point in passing it in. (This should probably also be `$overlay: $overlay`, if the widget parent class is fixed to accept an overlay.)

src/App.js
samwilson: It does seem strange that OO.ui.mixin.PopupElement doesn't handle the overlay. Anyway, given that it doesn't, should we fix that upstream, or just include something like this in the RevisionPopupWidget constructor?:

```js
this.$overlay = ( config.$overlay === true ? OO.ui.getDefaultOverlay() : config.$overlay ) || this.$element;
this.$overlay.append( this.popup.$element );
```

samwilson (changes requested):
Looks good and works in my testing. We just need to decide whether to add config.$overlay to the upstream PopupWidget mixin, or do it here ourselves (or both, so we can get on with things?).

src/App.js
mooeypoo: I am not sure if there's an issue adding this upstream; it might be that it's expected that if you have an overlay, you (the implementer) add it yourself, and that the ButtonPopupWidget does it for you as the "implementer of the popup".

But we should raise this and see if this can be upstreamed.

I like your idea of doing this in the config, but I think that if we append to the overlay consistently it's an okay solution for now anyways.

src/Controller.js
```diff
@@ -342,6 +343,21 @@ class Controller {
...
 	}
+
+	/**
+	 * Get the overlay div with the class wwt-overlay
+	 * Will add it if it doesn't already exist in the DOM.
+	 *
+	 * @return {jQuery} Overlay div
+	 */
+	getOverlay() {
+		if ( this.$overlay instanceof $ ) {
```
mooeypoo, line 354: I'm not a huge fan of 'get' methods that also make the widget/element, I prefer we create the element once on init and then attach it, and have the get method return the element we know exists already.

That said, I know that we do this for the getButton so while I am not a huge fan, I think consistency is okay here. I do think we might want to take a look in the future to make sure this doesn't create some annoyances like creating/attaching an overlay when it's not needed (like happened with the button), so maybe just keep our eyes open for a potential future bug.

mooeypoo (approved):
Some note-for-the-future comments, and we should try and upstream this. But for now, this looks good!

src/Controller.js
samwilson: Interesting. I actually feel the opposite, I think! :) It's also the same pattern that `OO.ui.getDefaultOverlay()` uses. It sort of makes sense to me, in that it ensures we only create and attach the thing when we know we need it (and we know we need it because someone called `getOverlay()`; no extra context required).

PR #122, opened 2019-12-13
Stop load if we are on an invalid page

mooeypoo:
This used to work, but when we changed the way we call the button,
it still got attached to the DOM even though the system in general
is inoperable.

This commit fixes it by adding the 'isValidPage' check before we
call the button and attach the event on it at all, so a button
isn't created and events aren't attached unnecessarily.

Bug: https://phabricator.wikimedia.org/T240690


samwilson (approved):
Works in my testing.

PR #123, opened 2019-12-16
Restore comment fading in, with jQuery fadeIn()

samwilson:
This does away with the custom fading in that we were using
before the comment was a jQuery object, and makes the fading-in
work again.

### DIFF
--- a/src/RevisionPopupWidget.js
+++ b/src/RevisionPopupWidget.js
@@ -74,10 +74,11 @@ function getCommentHtml( data, isCached ) {
 		$revCommentSpan.prepend( $( '<br>' ), $commentSpan );
 
 		if ( !isCached ) {
-			$commentSpan
-				.addClass( 'wwt-revisionPopupWidget-comment-animated' );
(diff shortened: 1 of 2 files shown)


MusikAnimal (commented):
I unilaterally used CSS in the beginning (even though we had jQuery), because it uses hardware acceleration, and distinguishes presentation vs interaction. But .fadeIn() is indeed much cleaner :) No concerns there.

I noticed that there is still a fade-in animation for cached data, when there wasn't before. Is this a problem?

MusikAnimal:
> I noticed that there is still a fade-in animation for cached data, when there wasn't before. Is this a problem?

Additionally, it looks like it shows "_()_" when you first open the popup containing cached data, then immediately changes to the full summary. This can cause a flickering effect if the size of the popup changes.

MusikAnimal:
I wait on this for right now. https://phabricator.wikimedia.org/T240707 will require reworking much of RevisionPopupWidget.

samwilson:
No longer required.

PR #124, opened 2019-12-17
Rework display of usernames and RevisionPopup

MusikAnimal:
Remove display of username/IP from infobar.

Fetch the username from the MediaWiki API along with the other revision
data.

Minor refactoring of RevisionPopupWidget, renaming some methods to say
"content" instead of "html" for clarity, since they are returning
jQuery objects.

Bug: https://phabricator.wikimedia.org/T240707

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - Prevent the WhoWroteThat button from appearing on the main page.
 
+### Changed
+- Fetch username from the MediaWiki API, and fade-in all content in the revision popup.
+
 ## [0.11.0] - 2019-12-13
 ### Added
 - This change log.
--- a/src/App.js
+++ b/src/App.js
@@ -175,17 +175,14 @@ class App {
 				if ( this.revisionPopup.isVisible() ) {
 					return;
 				}
-				const ids = this.getIdsFromElement( e.currentTarget ),
-					tokenInfo = wwtController.getApi().getTokenInfo( ids.tokenId );
+				const ids = this.getIdsFromElement( e.currentTarget );
 				this.activateSpans( $content, ids.editorId );
-				this.widget.setUsernameInfo( tokenInfo.username );
 			} )
 			.on( 'mouseleave', () => {
 				if ( this.revisionPopup.isVisible() ) {
 					return;
 				}
 				this.deactivateSpans( $content );
-				this.widget.clearUsernameInfo();
 			} )
 			.on( 'click', e => {
 				const ids = this.getIdsFromElement( e.currentTarget ),
@@ -193,17 +190,15 @@ class App {
 					isCached = wwtController.getApi().isCached( tokenInfo.revisionId );
 
 				this.activateSpans( $content, ids.editorId );
-				this.widget.setUsernameInfo( tokenInfo.username );
 				this.revisionPopup.show( tokenInfo, $( e.currentTarget ), isCached );
 				this.revisionPopup.once( 'toggle', () => {
 					this.deactivateSpans( $content );
-					this.widget.clearUsernameInfo();
 				} );
 
 				// eslint-disable-next-line one-var
 				const reqStartTime = Date.now();
-				// Fetch edit summary then re-render the popup.
-				wwtController.getApi().fetchEditSummary( tokenInfo.revisionId )
+				// Fetch revision data then re-render the popup.
+				wwtController.getApi().fetchRevisionData( tokenInfo.revisionId )
 					.then( successData => {
 						const delayTime = (
 							!isCached &&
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -32,7 +32,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Minor out-of-date code removed or updated
 - Change the sidebar link so it only works when clicked directly
   (and not when the empty space next to it is clicked)
-- Show revision size in popup even when there's no revision comment
+- Fetch username from the MediaWiki API, and fade-in all content in the revision popup.
```
samwilson: 0.11.0 has already been released, and shouldn't have it's changelog changed. New changes should go above in 'Unreleased'.

src/RevisionPopupWidget.js
```diff
@@ -125,14 +112,24 @@ function getUserLinksHtml( data ) {
...
+	const $shimmerSpan = $( '<span>' )
+		.addClass( 'wwt-shimmer wwt-shimmer-animation' );
+	return $( '<div>' )
+		.append( $shimmerSpan, $shimmerSpan.clone(), $shimmerSpan.clone() );
+}
+
+/**
+ * Get content for the popup, called once data has been retrieved from the API.
+ * @param {Object} data
+ * @param {boolean} isCached
```
samwilson: Param doesn't exist.

src/RevisionPopupWidget.js
```diff
@@ -146,11 +143,37 @@ RevisionPopupWidget.prototype.show = function ( data, $target, isCached ) {
...
+}
 
+/**
+ * Show the revision popup based on the given token data, above the given element.
+ * Note that the English namespaces will normalize to the wiki's local namespaces.
+ * @param {Object} data As returned by Api.prototype.getTokenInfo().
+ * @param {jQuery} $target Element the popup should be attached to.
+ * @param {boolean} [isCached] Whether the comment promise is available, cached.
+ */
+RevisionPopupWidget.prototype.show = function ( data, $target, isCached ) {
```
mooeypoo, line 158: I'm thinking it might be better at this point to use the fact that the request is a promise for this.

That is, instead of having ifs and elses here, we can have the widget request the API for the promise, if it's uncached, set pending state, and unset it when the data is available. From that moment on, the promise is immediately-resolved, so doing it over and over would mean that we don't show loading animation but still request the cached promise.

Does this make sense? It just seemed a bit clearer usage (widget asks API for promise and shows) rather than something else calls the API and then calls the popup show function. This part used to work logically when we expected a single html "dump" (with or without comment) and repeatedly called `show` from outside the system -- but there's no real need for this now that we have a promise, and that we need to show the entire data of that promise in this widget. 

Does this make sense? I am thinking the widget can basically be more standalone by doing its own process, and whatever else calls it just toggles it on and off without having to know anythinga bout whether the promise exists or is cached, etc.

MusikAnimal: I think so... and it sounds great! I was under some weird impression that the widgets shouldn't have knowledge of the API (like a view), and App was sort of the controller (gets data from API and gives it to the view) based on page interaction. If I'm following you right, I think all of [this logic](https://github.com/wikimedia/WhoWroteThat/blob/8f199808e92e616192cbbee1e3a4ddf9adb498e1/src/App.js#L198-L219) will need to be moved to RevisionPopupWidget as well? My only concern is this seems like a lot more refactoring. Maybe we could do it with a follow-up commit?

CHANGELOG.md
MusikAnimal: Haha, I thought I was under the Unreleased heading, and just one of the release notes were cancelled out by this one. Fixed!

src/RevisionPopupWidget.js
MusikAnimal: Fixed

src/RevisionPopupWidget.js
mooeypoo: Yeah API is a bit of a limbo state; usually, it is the Controller (in an MVC) that does it, then uses the API response to change the model -- and the widgets only listen to the model and invoke the controller.

So, we can send the widget the controller, and just ask for the API promise instead of having to refactor both the widget and the controller. Also, you're pointing out another logic weirdness where the App (somewhat the initialization-style-controller?) is calling the controller, which is a little meh, but might be okay if it's consistent about where we invoke what. It would probably be better for the controller itself to do that method, and then have `App` call that method.

So with all that said -- I do agree that that's quite a bit of a deeper change. I am good with doing this as a later refactor, but maybe we should document it with a ticket, just so we don't forget after the holidays? I think that beyond making things more readable and "proper" (arguably :D ) the refactor can prevent some potential sync issues that could happen.

Let's open another ticket, and I'll re-review this patch as-is without refactoring.

mooeypoo: (btw, unrelated to this change, but an addendum to my comment about the MVC -- part of the mismatch here is that while we have a Model, it's for the general system and not so much for the individual widget state. That might be something we could look at too, see if it makes sense to have the popup listen to the model for its data/state (am I in loading state, do I have a cached data, when the fetch happens, etc) rather than do that internal logic inside itself? It's not a must, but if we're already going to look at the way it works and simplify it, it might be a good point to revisit.)

src/App.js
```diff
@@ -213,14 +208,14 @@ class App {
 						Object.assign( tokenInfo, successData );
 
 						setTimeout( () => {
-							this.revisionPopup.show( tokenInfo, $( e.target ) );
+							this.revisionPopup.show( tokenInfo, $( e.target ), isCached );
```
mooeypoo, line 211: (Note a blocker for this, just more a comment in general)
This is one of the reasons I think the wwtController should do this internally; sending over the internal "isCached" state seems iffy to me.

This patch isn't the first thing that does it, but when we do refactor, we should take another look at how to do it without sending this internal state everywhere.

mooeypoo (approved):
lgtm and works; there's a followup for later to look into the over-arching behavior with the promises and controller that we can hash out in the (near?) future https://phabricator.wikimedia.org/T241004

PR #125, opened 2019-12-18
New Grunt task to check Firefox addons listing messages

samwilson:
This adds the checkListings[:beta] task, which retrieves the
extension's data from the Firefox store and compares the name
and description with what's in the local i18n files.

Bug: https://phabricator.wikimedia.org/T239840

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -5,13 +5,16 @@ module.exports = function Gruntfile( grunt ) {
 		langFiles = grunt.file
 			.expand( { filter: 'isFile', cwd: 'i18n' }, [ '*' ] )
 			.filter( f => f !== 'qqq.json' ),
+		langBlob = {},
 		/**
 		 * Generate a language JSON blob from the i18n translation files
 		 *
 		 * @return {Object} JSON object of all translations, keyed by language code
 		 */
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -242,6 +245,53 @@ module.exports = function Gruntfile( grunt ) {
...
+						listingLang = 'pt-PT';
+					} else if ( lang === 'pt-br' ) {
+						listingLang = 'pt-BR';
+					} else if ( lang === 'sv' ) {
+						listingLang = 'sv-SE';
+					} else if ( lang === 'nb' ) {
+						listingLang = 'nb-NO';
+					}
+					if ( json.name[ listingLang ] !== langBlob[ lang ][ nameMsg ] ) {
+						grunt.log.error( 'Listing does not match for ' + lang + ': ' + nameMsg );
```
mooeypoo: If you're already displaying an error, maybe differentiate between mismatch message for name and description in the output. It's probably obvious when you see it, but may be easier to clearly say which is which in the output?

samwilson: I've updated the output to make it clearer, and to include the correct message for easier copying-and-pasting.

PR #127, opened 2020-01-02
Flip the behavior of disabled elements' opacity on hover

mooeypoo:
Bug: https://phabricator.wikimedia.org/T240262


samwilson (approved):
This looks good to me.

I notice that the ToC is disabled, even though the links in it work correctly. That's not to the fault of this patch though.

PR #128, opened 2020-01-02
Refactor RevisionPopupWidget for MVC

mooeypoo:
We've changed the system to use MVC but didn't really touch the
widget, so it had a lot of its own logic. That logic didn't quite
fit the new features we want (and are about to tweak) like extending
some of the pending animation to more than the comment, or adding
more of the dynamic information within the promises and events.

This commit refactors the widget to listen to the model and respond
to refresh events, and have the click event trigger the controller
to utilize the API. It means, among other things, that --
- We don't need to deliver extra data like isCached between the
  MVC barriers,
- It should be a little more organized with some clear order of
  operations
- The popup doesn't constantly get rebuilt -- it builds its structure
  and only replaces the actual data when updated.

Bug: https://phabricator.wikimedia.org/T241004

### DIFF
--- a/src/Controller.js
+++ b/src/Controller.js
@@ -358,6 +358,53 @@ class Controller {
 		$( 'body' ).append( this.$overlay );
 		return this.$overlay;
 	}
+
+	/**
+	 * Set the token that is currently active.
+	 * Fetch the information that relates to the token, and update the model
+	 * accordingly.
+	 *
+	 * @param {string} tokenId [description]
+	 * @param {jQuery} $target [description]
+	 */
+	setActiveToken( tokenId, $target ) {
+		const tokenInfo = this.getApi().getTokenInfo( tokenId ),
+			reqStartTime = Date.now();
+
+		// Ensure the popup is attached to the visible content,
+		// which it otherwise wouldn't be for image thumbnails.
+		if ( $target.find( '.thumb' ).length ) {
+			$target = $target.find( '.thumb' );
+		}
+
+		if ( !this.getApi().isCached( tokenInfo.revisionId ) ) {
+			this.model.setCurrentToken( null, $target, 'pending' );
+		}
+		this.getApi().fetchRevisionData( tokenInfo.revisionId )
+			.then(
+				successData => {
+					const delayTime = (
+						!this.getApi().isCached( tokenInfo.revisionId ) &&
+						Date.now() - reqStartTime < 250
+					) ? 250 : 0;
+
+					Object.assign( tokenInfo, successData );
+
+					setTimeout( () => {
+						this.model.setCurrentToken( tokenInfo, $target, 'success' );
+					}, delayTime );
+				},
+				// Failure
+				() => {
+					// Silently fail. The revision info provided by WikiWho
+					// is still present, which is the important part,
+					// so we'll just show what we have and throw a console
+					// warning.
+					mw.log.warn( `WhoWroteThat failed to fetch data for revision with ID ${tokenInfo.revisionId}` );
+					this.model.setCurrentToken( null, $target, 'failure' );
+				}
+			);
+	}
 }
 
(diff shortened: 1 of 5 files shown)


src/RevisionPopupWidget.js
```diff
@@ -3,10 +3,23 @@ import Tools from './Tools';
...
+ * @param {Controller} controller WhoWroteThat controller
+ * @param {Model} model WhoWroteThat model
  */
-const RevisionPopupWidget = function RevisionPopupWidget() {
-	this.$popupContent = $( '<div>' )
+const RevisionPopupWidget = function RevisionPopupWidget( controller, model ) {
+	const $shimmerSpan = $( '<span>' )
+		.addClass( 'wwt-shimmer wwt-shimmer-animation' );
+
+	this.controller = controller;
```
samwilson: Is `this.controller` used anywhere? It doesn't look like it.

src/RevisionPopupWidget.js
```diff
@@ -112,81 +81,119 @@ function getUserLinksContent( data ) {
...
-}
+RevisionPopupWidget.prototype.initialize = function () {
+	this.diffSizeLabel = new OO.ui.LabelWidget( {
+		classes: [ 'mw-diff-bytes' ]
+	} );
+
+	this.commentLabel = new OO.ui.LabelWidget( {
+		classes: [
+			'comment',
+			'comment--without-parentheses'
```
samwilson: If this is initialised here with this class name, then empty comments still get parentheses. It was being done within a check e.g. `if ( data.comment !== '' ) { … }`. Although I guess the fix here isn't to not create this here, but rather to make sure it's removed when there's no comment?

samwilson (changes requested):
This looks good to me, the only thing I could find is the empty comments issue. Looks good to get more OOUI constructs into the popup.

src/RevisionPopupWidget.js
mooeypoo: omg **that's** why there are empty parentheses! I could not find those anywhere and it drove me nuts.

Noted! Will add them after the comment is there (and if there is a comment at all)

src/RevisionPopupWidget.js
mooeypoo: You're right; the first iteration, I had the widget itself invoke the fetch action when it toggled (which MVC dictated it needs to call the controller)... then, after a bit more thinking, I thought it might be easier to make the widget listen to the model which meant that the controller is invoked within the `app`

I have to admit, the widget listening to the model and the model sending a sort of "set token" event is okay in principle, but I'm not 100% in love with it. I do think it separates concerns much better, which is good, but maybe the model should actually retain the data/store it and do something with it rather than "just" be a conduit for the event.

... there was no reason for it for now, but it might be something to take into account.  For example, if the top bar still displayed the nickname, it might be good to have that data in the model (set properly, event, etc) so the entire widget and system are synchronized. Since that doesn't happen anyways, the only thing that needs to know what was clicked was app (for "sticking" the highlight, which is UI-only) and the revision popup (thorugh the model) I thought it was okay to make this a temporary "notification" event.
We might want to look at this again in the future, though.
No biggie, just pointing this out.

I'll remove `this.controller` from the widget since it does no actual action and only listens to state events from the model.

src/Controller.js
```diff
@@ -345,11 +345,11 @@ class Controller {
 	}
 
 	/**
-	 * Get the overlay div with the class wwt-overlay
-	 * Will add it if it doesn't already exist in the DOM.
-	 *
-	 * @return {jQuery} Overlay div
-	 */
+	* Get the overlay div with the class wwt-overlay
```
samwilson: Whitespace went funny here.

samwilson (approved):
Tiny formatting thing; I'll fix it up now and merge.

PR #132, opened 2020-01-08
Normalize lang code in model to lowercase

samwilson:
Setting a language via ?uselang=pt-br results in lang="pt-BR" being added to the HTML, where WWT reads it from. This change normalizes this to match the letter case of the translation files, so when we load them in the Controller they are correctly in the translations object.

https://phabricator.wikimedia.org/T239634

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -41,7 +41,7 @@ class Model extends EventEmitter {
 	 */
 	initialize( $content, config ) {
 		this.$contentWrapper = $content;
-		this.lang = config.lang || 'en';
+		this.lang = config.lang ? config.lang.toLowerCase() : 'en';
 		this.namespace = config.namespace || '';
 		this.mainPage = !!config.mainPage;
 		this.revisionId = config.revisionId || 0;


src/Model.js
```diff
@@ -41,7 +41,7 @@ class Model extends EventEmitter {
 	 */
 	initialize( $content, config ) {
 		this.$contentWrapper = $content;
-		this.lang = config.lang || 'en';
+		this.lang = config.lang ? config.lang.toLowerCase() : 'en';
```
mooeypoo, line 44: I'm kinda wondering if we shouldn't do this lowerCasing before ... like, when we instantiate the controller in `browserextension.js` line #90 ...

samwilson: I wondered about that too, but ["be liberal in what you accept"](https://en.wikipedia.org/wiki/Robustness_principle) and all that... if everywhere reads the lang from the model (which it does) then it's when setting it in the model that we should make sure it's of correct form.

I should of course have added a test!

mooeypoo: Hm I guess it could be argued either way; the model could also be used to store whatever is given to it, being "reusable in other systems" but that is quite heavy bikeshedding here :P 

I'm good with this.

PR #136, opened 2020-01-13
Keep content attached to the DOM when WWT is active

samwilson:
Rather than removing the original HTML content, and keeping a deep
clone in memory for reattaching later, this puts the 'cache' of
the original content into a hidden div that's kept attached to
the DOM. This means that all events and other behaviour is
maintained, and works better for some misbehaving gadgets.

Bug: T231424

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -57,7 +57,14 @@ class Model extends EventEmitter {
 	}
 
 	/**
-	 * Cache the original content we are about to replace
+	 * Cache the original content we are about to replace.
+	 *
+	 * This is done by appending it to the DOM in a hidden div, a slightly hacky workaround to
+	 * accommodate Wikipedia gadgets that do weird things. Normally, deep jQuery-cloning an element
+	 * means it can be re-attached to the DOM later and everything will keep working, but there are
+	 * a few situations (e.g. mis-nested elements; see T231424 for details) where the resulting DOM
+	 * doesn't match what was originally there. To avoid that, we keep the original content attached
+	 * and just hide.
 	 *
 	 * @param {jQuery} [$content=this.$contentWrapper.contents()] A jQuery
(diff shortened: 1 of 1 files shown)


src/Model.js
```diff
@@ -66,7 +73,17 @@ class Model extends EventEmitter {
 	 *  it is replaced.
 	 */
 	cacheOriginal( $content = this.$contentWrapper.contents() ) {
-		this.$originalContent = $content.clone( true, true );
+		const cacheClass = 'wwt-cached-original-content';
+		this.$originalContent = $content;
+		let $cache = $( '.' + cacheClass );
+		// If the cache element doesn't already exist, add it.
+		// It will then be re-used for subsequent re-activations of WWT.
+		if ( $cache.length === 0 ) {
```
mooeypoo, line 81: I didn't test this yet but we need to be careful here; we change the "original content" after VE save, so this should make sure that allows the replacement.

I think maybe check the cache dom exists but also that the revid didn't change? If it did change, replace the content, and we will need to re run the event attachments.

mooeypoo (changes requested):

src/Model.js
samwilson: it works correctly in my tests with VE: running WWT, making an edit with VE, and then running WWT again (including multiple times on and off) shows the right HTML.

mooeypoo: I'm actually confused as to how it is working, if you're skipping re-assigning the DOM if the $cache already has it O.O but I trust your tests.

mooeypoo: Oh, I see why -- this is only about creating and appending the DOM; it should be replaced automatically by the fact it has the mediawiki content classes with VE events -- and then not re-appended.

Taking my concern back, I'm a lot less confused now :D

mooeypoo (approved):

PR #137, opened 2020-01-13
Use wgUserLanguage instead of the html lang attribute

samwilson:
This also reverts the workaround that I added in #132 because it's
no longer required.

Bug: T239634


samwilson:
@mooeypoo I was wrong yesterday; `mw.config.get( 'wgUserLanguage' )` works fine for anons. So I think this is good to go.

PR #138, opened 2020-01-13
Correct display of revision popup content

mooeypoo:
After the refactor and especially after we've changed the way
that the content loads in the popup (loading usernames from the
mediawiki API rather than showing them outright while other content
loads, etc) the display and structure of the popup needed to be
readjusted.

In this commit:
- Reorganize the LESS structure so all css classes are scoped for
  within the popup, and do not leak for mw classes outside of it.
  (Fixes bug https://phabricator.wikimedia.org/T241998)
- Wrap long usernames (and IPv6 usernames)
- Add spacing after diff numbers, but only if there is a comment
- Bring back the italicization of the comment.
- Place the size diff at the intro line if there's no comment,
  or near the comment if that comment exists.
- Set padding at 0.75em and 0.5em for the popup (see T242954)

Bug: https://phabricator.wikimedia.org/T241004
Bug: https://phabricator.wikimedia.org/T242954

### DIFF
--- a/src/RevisionPopupWidget.js
+++ b/src/RevisionPopupWidget.js
@@ -22,7 +22,6 @@ const RevisionPopupWidget = function RevisionPopupWidget( model, config = {} ) {
 	// Parent constructor
 	RevisionPopupWidget.parent.call( this, Object.assign(
 		{
-			padded: true,
 			autoClose: true,
 			position: 'above',
 			// FIXME: 'force-left' for RTL languages
@@ -35,6 +34,8 @@ const RevisionPopupWidget = function RevisionPopupWidget( model, config = {} ) {
 
 	this.initialize();
 	this.model.on( 'setToken', this.onModelSetToken.bind( this ) );
+	this.$element
+		.addClass( 'wwt-revisionPopupWidget' );
 };
 
 /* Setup */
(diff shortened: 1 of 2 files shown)


MusikAnimal (changes requested):
Maybe I'm doing something wrong, but I still see summaries being cut off and some weird spacing, e.g. at https://en.wikipedia.org/wiki/Mast_(sailing):

Screenshot from 2020-01-13 23-55-39

Screenshot from 2020-01-13 23-57-06

mooeypoo:
I missed the word break rule for the summaries. Good catch.

And looks like the percentage row label should be `display: block` so it's always on its own line.

mooeypoo:
Good catch on all; I fixed up the widget again and the LESS rules. I also added the requested padding from https://phabricator.wikimedia.org/T242954 so this PR should be fixing both tasks.

MusikAnimal:
Looks awesome! There's some extra spacing before the diff size now, that wasn't there before.

Before:

Screenshot from 2020-01-21 12-15-46

After:

Screenshot from 2020-01-21 12-14-54

Does this matter? Before it was using a space, now it's using `margin-left: 0.5em`. A space would match the appearance in revision histories:

Screenshot from 2020-01-21 12-17-32

mooeypoo:
Yeah, we can't really bring back the `&nbsp;` with the label now that it's an OOUI widget rather than plaintext, and honestly we should be doing it all with proper CSS even if MediaWiki core doesn't, especially since there are slightly different rules for when there is a comment vs not.

I can decrease the spacing to make it a little more similar, but I'd ping @prtksxna here to see what type of spacing he'd like? We can also always iterate on it.

PR #139, opened 2020-01-13
Add message when edit summary is hidden

cimurah:
Add "edit summary removed" message when the edit summary has been hidden in the revision
details pop-up.

Bug: T241062

### DIFF
--- a/src/RevisionPopupWidget.js
+++ b/src/RevisionPopupWidget.js
@@ -162,7 +162,13 @@ RevisionPopupWidget.prototype.updateData = function ( data = {} ) {
 	// Comment content
 	this.commentLabel.setLabel(
 		new OO.ui.HtmlSnippet(
-			Tools.bidiIsolate( $.parseHTML( data.comment ) )
+			data.comment === undefined ?
+				$( '<span>' )
+					.addClass( 'history-deleted' )
+					// Note we can't use the native MediaWiki 'revdelete-summary-hid'
+					// message because it can include parser functions.
+					.text( mw.msg( 'whowrotethat-revision-edit-summary-hidden' ) ) :
+				Tools.bidiIsolate( $.parseHTML( data.comment ) )
 		)
 	);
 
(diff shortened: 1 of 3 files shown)


src/RevisionPopupWidget.js
```diff
@@ -162,7 +162,11 @@ RevisionPopupWidget.prototype.updateData = function ( data = {} ) {
 	// Comment content
 	this.commentLabel.setLabel(
 		new OO.ui.HtmlSnippet(
-			Tools.bidiIsolate( $.parseHTML( data.comment ) )
+			data.comment === undefined ?
+				$( '<span>' )
+					.addClass( 'history-deleted' )
+					.text( mw.msg( 'whowrotethat-revision-edit-summary-hidden' ) ) :
```
samwilson, line 170: Can the MediaWiki message `revdelete-summary-hid` be used here instead? To reduce duplication and translation efforts.

cimurah: I noticed comment https://github.com/wikimedia/WhoWroteThat/blob/c58d94ca7e54a50c21bea62e50822f5c5a518869/src/RevisionPopupWidget.js#L58. Would using `revdelete-summary-hid` create  the same concern regarding using a native Wikimedia message that might include parser functions?

cimurah: I noticed comment: 
https://github.com/wikimedia/WhoWroteThat/blob/c58d94ca7e54a50c21bea62e50822f5c5a518869/src/RevisionPopupWidget.js#L58
Would using `revdelete-summary-hid` create the same concern regarding using a native MediaWiki message that might include parser functions? If so, should I also add that comment here?

samwilson: Ah, that makes sense. In that case, should we add `{{msg-mw|revdelete-summary-hid}}` to the qqq docs to highlight the fact that it's the same message?

cimurah: I added the comment to qqq docs and also to the code just in case someone is only looking at the code.

samwilson (approved):
Looks good. Works in my testing.

samwilson:
Oh oops, the changelog also needs to be updated!

PR #145, opened 2020-01-16
Add mw.Api to initial load dependencies

mooeypoo:
While initializing the controller, we're loading the WWT Api class
with an instance of mw.Api(); the initialization of the controller
happens before we run `loadDependencies` (which runs on 'launch'
because it loads all of OOUI modules we need).

While it's unclear how this didn't break before (!?) it is safer
to add `mediawiki.Api` to the initial dependencies in the RLQ loader.


MusikAnimal:
Does this warrant a CHANGELOG entry?

mooeypoo:
Good point! Adding

aezell (approved):
No idea how this worked before but it's correct now.

PR #150, opened 2020-01-22
Update dependencies and silence eslint error

samwilson:


samwilson:
This eslint error came about because of the order we merged #143 and #136.

PR #159, opened 2020-02-04
Update text for contact error

cimurah:
Change the language in "whowrotethat-error-contact" to
"Error: Refresh or try again later. If the issue persists, contact us."
so that it is more inclusive and general for all cases.

Bug: T24331


i18n/en.json
```diff
@@ -12,10 +12,10 @@
 	"whowrotethat-state-pending": "$1 is loading. This might take a while.",
 	"whowrotethat-state-pending-name": "Who Wrote That?",
 	"whowrotethat-state-error": "Error: $1",
-	"whowrotethat-error-refresh": "Please refresh the page or try again later.",
+	"whowrotethat-error-refresh": "Refresh or try again later.",
 	"whowrotethat-error-later": "Please try again later.",
-	"whowrotethat-error-contact": "Please $1 about this issue.",
-	"whowrotethat-error-contact-link": "contact us",
+	"whowrotethat-error-contact": "Refresh or try again later. $1",
```
samwilson, line 17: Should the 2nd sentence be removed from this message and just appended after it, now that the link isn't inside a sentence? That'd mean we could avoid the `Tools.i18nHtml()` usage as well.

cimurah: Appending the two sentences would be ideal, but since the second sentence is a link it gets a bit trickier to append a string and the jQuery object (link) so we need to use `Tools.i18nHtml()`.  We would need to create another message that outputs an error that contains the two sentences separately since the 2nd sentence is a link if we want to avoid using `Tools.i18nHtml()`.  I think this approach makes it easier for translators to know that the two sentences go together.
Please let me know what you think. Thank you!

samwilson: Sure, that makes sense. Merging now.

PR #165, opened 2020-06-06
Remove language barrier and make it applicable to any wikipedia language

atsushieno:
This extension targets only a few languages in Wikipedia, which is simply not acceptable in terms of language diversity.

I am not sure if this builds as the build script is broken for me, so I'll leave it to the CircleCI build.

The comprehensive list is generated by simple sed on mediawiki sources:

- `grep -nR "NS_SPECIAL.*=>" languages/messages/* | sed -re "s/.*Messages([A-Za-z_]+).php:.*NS_SPECIAL.*=>.*['\"](.*)['\"],/\t\t\t\"\*\:\/\/\L\1.wikipedia.org\/wiki\/\2:*\",/"` for `exclude_globs`
- `grep -nR "NS_SPECIAL.*=>" languages/messages/* | sed -re "s/.*Messages([A-Za-z_]+).php:.*NS_SPECIAL.*=>.*['\"](.*)['\"],/\t\t\t\"\*\:\/\/\L\1.wikipedia.org\/\*\",/"` for `matches`


mooeypoo:
Unfortunately, the WhoWroteThat extension does not control the language distribution. 

WhoWroteThat shows the revision history data analysis from [WikiWho service](https://api.wikiwho.net/en/whocolor/v1.0.0-beta/) that's operated by the [Leibnitz Institute for the Social Sciences](https://www.gesis.org/en/institute/departments/computational-social-science/). Their algorithm is currently limited on the number of language-Wikipedias it is running on. We have been in contact with them to increase the number of Wikipedia languages they support, and there may be some more coming up, but they are definitely not supporting all ~380 content languages that Wikipedia currently runs.

WhoWroteThat extension's language setting is not there to create a barrier, but rather to make sure the extension only operates only on Wikipedias where we are able to provide the data. If we take that ruleset off (or expand it in) our config file, the extension will load on Wikipedias but will not be able to provide the functionality, which is unhelpful and confusing to users.

For more reference, please see the [product help page on MediaWiki.org](https://www.mediawiki.org/wiki/Who_Wrote_That%3F#Current_Limitations_of_the_Tool).

atsushieno:
Fair enough. Thanks for clarification.

PR #166, opened 2020-07-21
Update ESLint and other linters

edg2s:
* Fix documentation whitespace
* Avoids use of 'var'
* Other minor fixes


src/Tools.js
```diff
@@ -19,7 +19,7 @@ class Tools {
 	 * @param {Mixed} [params] Parameters to output into the console
 	 *  alongside the message. This is especially useful for errors.
 	 */
-	static log( str, type = 'debug', params ) {
+	static log( str, type = 'debug', params = undefined ) {
```
mooeypoo, line 22: (this isn't a merge blocker, just more a curiosity question) 

Is that not a given? I don't mind it in the code, I just didn't know anything expects this? Is that not automatically true and therefore unnecessarily stated? Kinda like doing `let x = undefined;` when really `let x;` means the same and is enough...? Am I missing something?

test/.eslintrc.json
```diff
@@ -0,0 +1,9 @@
+{
+	"extends": [
+		"wikimedia/mocha"
```
mooeypoo, line 3: I didn't know we *have* a mocha ESlint declaration! Is that new?

edg2s: Yeah, about 2 months. It's also used by the wikimedia/selenium config, which is more common.

src/Tools.js
edg2s: Yes, but the linter complains that the there are non-default arguments after defaulting ones. It just so happens that in JS there is an implicit default. Many other languages would not allow the code as written before (PHP, Java..)

edg2s: I would probably rewrite this as allowing variadic additional arguments, but that's outside the scope of this commit.

mooeypoo: oh wow, I didn't even notice the non-default after default params, that's a good point.

What do you mean as allowing variadic additional arguments? Doing something like `var args = Array.prototype.slice.call(arguments);` without representing it in the actual signature? 

(We could totally do that, I think this above one was mostly as a quickfix to make sure we are always outputting something sensible to the console, but that would mean also fixing up the calls, etc.) Is it really better? Or just a style preference?

edg2s: Yeah. console.log etc are variadic, so log should probably be as well:

`log( 'foo', 'info', 1, 2, 3)` becomes `console.info( 'Prefix: foo', 1, 2, 3 )`

PR #167, opened 2020-07-21
Fetch messages when popup is requested instead of when extension is loaded

edg2s:
Bug: [T258457](https://phabricator.wikimedia.org/T258457)


MusikAnimal (approved):
This only took what, 2 years to review... very sorry about the delay!

Anyways this works great in my testing. Thanks so much for volunteering your time :)

PR #168, opened 2020-07-21
Fix padding on narrow screen vector

edg2s:
Below 982px Vector uses 1em content padding instead of 1.5em


edg2s:
Before fix:
image
After fix:
image

PR #170, opened 2022-06-28
Remove unused 'activeTab' permission, update Node dependencies

MusikAnimal:
This commit also ignores a stylelint violation that occurred after
upgrading stylelint-config-wikimedia. The font size in this case was
explicitly given by the designer, see T234711.

Bug: [T311581](http://phabricator.wikimedia.org/T311581)


samwilson (approved):
Looks good to me! Everything compiles fine, but I'm failing to test because I've forgotten how to get Firefox to install unsigned extensions. :-P

The only thing I think might need changing is that the version number in package.json should be bumped. But that can happen in a separate patch.

MusikAnimal:
> Looks good to me! Everything compiles fine, but I'm failing to test because I've forgotten how to get Firefox to install unsigned extensions. :-P

It looks like the README needs to be updated. To install manually, nowadays you go to about:debugging#/runtime/this-firefox, but the `grunt run` command using web-ext should also work as described.

> The only thing I think might need changing is that the version number in package.json should be bumped. But that can happen in a separate patch.

Yeah I was going to save that for a "version bump" PR along with updates to CHANGELOG, etc.

samwilson:
Sounds good.

Thanks for working on this!

PR #172, opened 2022-07-29
Fix a bug with pages that have multiple .mw-parser-output

snapwich:
right now when this app replaces the page content (when enabling the plugin on a wikipedia page in the left nav) it searches for a `.mw-parser-output` class. usually there is only one of those classes present on the page around the primary page contents, however for pages that have `.mw-indicators` (icons for things such as "this page is protected") those icons are also wrapped in `.mw-parser-output`. the app currently treats every `.mw-parser-output` as a content wrapper so if you have two indicators when the page is re-appended to the wrapper it will write three copies of the page contents to the page.

example page with this bug: https://en.wikipedia.org/wiki/Bill_Gates

a side-effect of this bug (besides having the page display multiple times) is that the hovering effects will only apply to the last copy of the page contents. so if the page contents are written three times you need to scroll to the last copy of the page contents for the hover effects to work.

### DIFF
--- a/src/Controller.js
+++ b/src/Controller.js
@@ -144,7 +144,7 @@ class Controller {
 				// This is due to VE replacing the content element,
 				// rather than inserting new contents into it.
 				// eslint-disable-next-line no-jquery/no-global-selector
-				this.model.initialize( $( '.mw-parser-output' ), config );
+				this.model.initialize( $( '#bodyContent .mw-parser-output' ), config );
 				this.model.toggleEnabled( true );
 			} );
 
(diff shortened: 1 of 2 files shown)


MusikAnimal:
Wow, I'm quite surprised this has never come up until now... thank you for identifying this issue and creating a PR! I'm testing this now

MusikAnimal:
> I'm testing this now

So my computer froze and I had to restart, then completely forgot what I was doing after restarting, hehe!

Anyway, now having tested this, it fixes the bug as far as I can tell and I see no other places where `mw-parser-output` needs to be udpated (in comments or otherwise). Thanks again for submitting this PR!

MusikAnimal:
We'll get a new release of the browser extension out ASAP.

PR #173, opened 2022-07-30
Prepare 0.16.1 release

MusikAnimal:


samwilson (approved):
(I didn't know you could request review from the whole team!)

This looks fine. Only thing is, the previous release (0.16.0) seems to be missing: https://github.com/wikimedia/WhoWroteThat/releases

MusikAnimal:
> (I didn't know you could request review from the whole team!)

I didn't either, but it pings everyone and not just the devs, so I guess we probably won't use it much.

> the previous release (0.16.0) seems to be missing: https://github.com/wikimedia/WhoWroteThat/releases

Whoops! Looks like I forgot to make a release last time :( I just created it. It's got the wrong date but it is tied to the right commit, at least.

With your approval I'll go ahead and merge, and this time I won't forget to make the release!

PR #175, opened 2022-09-14
Prepare 0.16.2 release

MusikAnimal:
Also add dev step to use Node version specified in .nvmrc


theresnotime (approved):

snapwich:
these releases are making it to [firefox store](https://addons.mozilla.org/firefox/addon/whowrotethat) but don't seem to make it to [chrome](https://chrome.google.com/webstore/detail/who-wrote-that/ekkbnedhfelfaidbpaedaecjiokkionn). do you know when they'll make it to the chrome store? thanks

PR #179, opened 2023-05-04
Add pt.wikipedia.org

outl2unt123:


outl2unt123:
@MusikAnimal

MusikAnimal:
Thanks for this! I will get a few more of the newly added languages added to WhoWroteThat too, and we'll make a new release this week :)

PR #181, opened 2023-06-18
Fix Vector 2022 sticky header overlapping info bar

thewub:
Fix Vector 2022 sticky header overlapping the info bar following instructions at https://www.mediawiki.org/wiki/Reading/Web/Desktop_Improvements/Features/Sticky_Header#My_templates_use_sticky_elements._How_do_make_sure_they_do_not_overlap_the_sticky_header?

Bug: T310424


MusikAnimal (approved):
Works great in my testing. Thanks for contributing!

PR #182, opened 2023-10-17
Fix typo that loads messages when popup is requested

cimurah:
Messages were not being loaded and rendered as expected due to a typo.

Bug: T334766


MusikAnimal (approved):
Nice catch! This has been broken for 3+ years... 51b0b3a9c831259880d5b605f3ffde3e2f5a0d18

PR #184, opened 2024-02-27
Update Node version and dependencies, fix linting errors

MusikAnimal:
Remove the obsolete wikiWhoProxy.php; we now host our own installation


MusikAnimal:
Might as well review this and the bug fix together. Closing in favor of #185

PR #185, opened 2024-02-27
Fix CSS selector for .mw-parser-output to work in Vector 2022

MusikAnimal:
Vector 2022 puts indicators in the article body, so they also have the `.mw-parser-output` CSS class. This commit changes it to look only within `#mw-body-text`, which should work across all skins.

Bug: https://phabricator.wikimedia.org/T320364


MusikAnimal:
The second commit is the bug fix. The first just updates dependencies.

PR #188, opened 2024-04-01
Encode article title in WikiWho API access

ChlodAlejandro:
The article title is not URL encoded when the WikiWho API URL is being constructed ([Api.js#getAjaxURL](https://github.com/wikimedia/WhoWroteThat/blob/b46af9f/src/Api.js#L129)). This causes any article with a question mark to have a URL containing query parameters starting wherever the question mark is.

Most prominent issue is when viewing an article with a title that ends with a question mark, where the title without the question mark removed (practically always a redirect) shows up instead of the actual article.
| **Before** | **After** |
:---:|:---:
image | Screen Shot 2024-04-01 at 23 40 37

Revision ID left unencoded since it's expected to be a number, which won't have any special characters that need encoding.

*Initially reported by [Your Power](https://en.wikipedia.org/wiki/User:Your_Power) on Discord.*

### DIFF
--- a/src/Api.js
+++ b/src/Api.js
@@ -126,7 +126,7 @@ class Api {
 			this.url,
 			subdomain,
 			'whocolor/v1.0.0-beta',
-			this.mwConfig.get( 'wgPageName' ),
+			encodeURIComponent( this.mwConfig.get( 'wgPageName' ) ),
 			// Always include the revision ID, to make sure we are always asking for
 			// the correct revision, whether the page was just edited, or, whether the
 			// page was edited by someone else while we were looking at the current page
--- a/test/suite/Api.test.js
+++ b/test/suite/Api.test.js
@@ -12,7 +12,7 @@ describe( 'Api test', () => {
 					wgRevisionId: 234,
 					wgPageName: 'Iñtërnâtiônàlizætiøn_(disambig)'
 				},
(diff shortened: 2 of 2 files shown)


ChlodAlejandro:
Looks like I forgot to fix a test (and write a new one). 😔 I'll do this in the morning (~8 hours), if it hasn't been resolved by then.

ChlodAlejandro:
Should be fixed now. I've ensured that the test [here](https://github.com/wikimedia/WhoWroteThat/pull/188/commits/dfb841a372e2fd21614395cedb88bfa877cd5e2c#diff-f5a5baf9418cd3b027fb3b7d5e18f7b7314ee86dea396cdab6cad79f78c8943aR15) now uses encoded paths and also added a [test to check if special characters are encoded](https://github.com/wikimedia/WhoWroteThat/pull/188/commits/dfb841a372e2fd21614395cedb88bfa877cd5e2c#diff-f5a5baf9418cd3b027fb3b7d5e18f7b7314ee86dea396cdab6cad79f78c8943aR37).

MusikAnimal (approved):
Fantastic! I'm really surprised this bug hasn't surfaced until now, hehe. Thanks for fixing it, and also for a adding a test!  :100:

We'll cut a new release and get this shipped soon, if not tonight (my time) then tomorrow.

